PR #66, opened 2015-02-11
Analysis update

bjornstenqvist:
Added quadrupole-analysis in DipoleAnalysis which now is named MultipoleAnalysis.
Ewald summation test works but drift is still present in simulations.


include/faunus/multipole.h
```diff
@@ -783,473 +780,514 @@ namespace Faunus {
...
+          o << "Dipole-dipole, lB=" << _lB << textio::_angstrom;
+          return o.str();          
+        }
+      protected:
+        double _lB;
+      public:
+        DipoleDipole(InputMap &in, const string &dir="") {
+          name="Dipole-dipole";
+          in.cd ( jsondir+"/coulomb" );
+          pc::setT(in("temperature",298.15 ));
```
mlund: Temperature no longer needs to be set from pairpotential

src/examples/keesom.cpp
```diff
@@ -15,7 +16,7 @@ int main() {
   auto b=a;                    // duplicate particle...
   b.x() = 0.9_angstrom;        // ...and separate by 0.9 angstrom
 
-  Potential::DipoleDipole u(298.15_K, 80.); // pair potential
```
mlund: For a simple example it may be instructive to have a constructor that takes basic input like temp. and diel. In this case, it also shows how to use units.

PR #85, opened 2016-12-13
energy matrix

mlund:


include/faunus/space.h
```diff
@@ -20,6 +20,15 @@
 namespace Faunus
 {
 
+    /**
+     * @brief The MoleculeVec class - used to identify a list of molecules
+     */
+    template<class T>
+    class FaunusVector : public std::vector<T> {
```
mlund: What is a "FaunusVector"?

include/faunus/space.h
```diff
@@ -79,6 +96,16 @@ namespace Faunus
           return sum;
       }
 
+      Group* get(int i)
```
mlund: To get the i'th group, wouldn't one just call `Space::groupList()[i]`? If this is not what this function is doing, the name is confusing.

include/faunus/space.h
```diff
@@ -205,6 +232,31 @@ namespace Faunus
 
   };
 
+  class TrackerGroup : public Tracker<Group *> {
```
mlund: Please add documentation for this. Do `_map` and `myTid` need to be public?

include/faunus/group.h
```diff
@@ -22,8 +22,16 @@ namespace Faunus
   {
   private:
       int molsize; // Number of atoms per molecule
+
+
   public:
       typedef PropertyBase::Tid Tid;
+      int molIndex = -1; /// \brief Index of group in molecular tracker (sum of previous molecules in vector and on lower TID)
```
mlund: I'm not sure `group` should be responsible for knowing where it is located in `Space`. This adds an interdependency that should be avoided. Is the problem that you in energy functions get a group and not an index?

include/faunus/move.h
```diff
@@ -2554,39 +2548,14 @@ namespace Faunus
...
-    /**
+    /**pot
      * @todo Early rejection could be implemented
      *       - not relevant for geometries with periodicity, though.
      */
     template<class Tspace>
     double Isobaric<Tspace>::_energyChange()
     {
-        double uold = _energy(spc->p);
+        double uold = ( dp < 1e-6 ) ? 0.0 : pot->systemEnergy(spc->p); // calculates internal energy of all molecules, redundant but neccesary for Emat
```
mlund: `if`-statements are generally easier to read

include/faunus/move.h
```diff
@@ -2596,7 +2565,7 @@ namespace Faunus
             for ( auto i : *g )
                 if ( spc->geo.collision(spc->trial[i], spc->trial[i].radius, Geometry::Geometrybase::BOUNDARY))
                     return pc::infty;
-        double unew = _energy(spc->trial);
+        double unew = ( dp < 1e-6 ) ? 0.0 : pot->systemEnergy(spc->trial);
```
mlund: ditto

include/faunus/space.h
```diff
@@ -57,9 +66,11 @@ namespace Faunus
   template<class T, class Tid=int>
   class Tracker
   {
-  private:
-      std::map<Tid, std::vector<T> > _map;
+  protected:
```
mlund: why not private?

include/faunus/space.h
```diff
@@ -316,6 +368,7 @@ namespace Faunus
           for ( auto mol : molecule )
               while ( mol.Ninit-- > 0 )
                   insert(mol.id, mol.getRandomConformation(geo, p));
+          molTrack.fillIndex();
```
mlund: Could you not just use the original tracker and here fill it without calling a member function?

include/faunus/analysis.h
```diff
@@ -2409,12 +2409,12 @@ namespace Faunus
 
             spc->geo.setVolume(Vnew);
             pot->setSpace(*spc);
-        }
+          }
```
mlund: bad indent

include/faunus/group.h
```diff
@@ -22,8 +22,16 @@ namespace Faunus
...
+      int Index;        /// \brief Index of group in molecular tracker sub vector
```
mlund: variables start with small letters

include/faunus/energy.h
```diff
@@ -36,6 +37,8 @@ namespace Faunus
   namespace Energy
   {
 
+
```
mlund: lot of space

include/faunus/energy.h
```diff
@@ -139,7 +142,7 @@ namespace Faunus
         { return 0; }
 
         /** @brief Update energy function due to Change */
-        virtual double updateChange( const typename Tspace::Change &c ) { return 0; }
+        virtual double updateChange(bool acc) { return 0; }
```
mlund: Why delete the old function? It's meant to be used.

include/faunus/energy.h
```diff
@@ -2870,9 +3286,19 @@ return u * conc; // -> kT
...
             for ( int j = i + 1; j < (int) spc.groupList().size(); j++ )
                 u += pot.g2g(p, *spc.groupList()[i], *spc.groupList()[j]);
+
+        if(u != systemEnergy2(spc, pot, p))
+            std::cout << "systemEnergy != systemEnergy2, " << u << " " << systemEnergy2(spc, pot, p) << endl;
         return u;
     }
 
+    template<class Tspace, class Tenergy, class Tpvec>
+    double systemEnergy2( Tspace &spc, Tenergy &pot, const Tpvec &p )
```
mlund: Why do you need to make a `systemEnergy2()`? Could you not just add `setSpace()` to the original `systemEnergy()`?

include/faunus/energy.h
```diff
@@ -2921,6 +3347,154 @@ return u * conc; // -> kT
         return du;
     }
 
+    template<class T1, class T2>
```
mlund: Add documentation for this

include/faunus/energy.h
```diff
@@ -151,8 +154,408 @@ namespace Faunus
...
+                    if ( unew == pc::infty )
+                        return pc::infty;   // early rejection
+                }
+            }
+            return unew;
+        }
     };
 
+
+class EnergyMatrix {
```
mlund: Have this in a separate header? If you plan to make use the same code here and in other programs that would make it easier to update. Documentation would be nice.

mlund:
@lukassukenik

mlund:
Thanks, Lukas -- merge seems possible, but the build or unittests run via travis seem to fail, follow link above. Any ideas?

LukasSukenik:
Yes, I didnt commit changes to geometry.cpp, where i instantiate the static
member next_id from group class.

The practice is to do it after the definition of class, however it needs to
be in a *.cpp file to be used only once. Hence i put it there. It may not
be ideal placement and probably should be moved elsewhere. I will leave it
to your judgement, Maybe a group.cpp file?

mlund:
I'd put the static initialization in a separate `group.cpp` file for clarity (must also be added to `src/CMakeLists.txt`. Comments:

1. You increment group's static variable upon every construction to automatically id all groups 0, 1, 2, etc. which should reflect the position in the Space vector of groups. What if someone in between generates a temporary group? This is not at all an unlikely scenario. One could instead directly assign `id` after adding to the vector in Space. In any case, this would not work for GC, but that's OK for the moment.

2. In `TranslateRotate` you went back to the old way of calculating energies, i.e. not through the `change` object. This will break Ewald summation. Is there a problem in using the change object?

LukasSukenik:
ad 1) No, its purely for identification and unordered map is then used in
EnergyMatrix class to "map" id to "energy matrix index", There is no need
to for the groups to be ordered as they are stored in molTracker/GroupList.
Its more robust this way since no molTracker information is neccesary.

ad 2) Energy matrix now uses change object, Hopefully correctly( make test
passed )

I changed the implementation to use change object. And will be doing a pull
request shortly.

However I made a slight alteration to the energyChangeConfiguration()
function to use g2All function:

      template<class Tenergy, class Tpvec, class Tgeo, class Tparticle>

      double energyChangeConfiguration (

          Space<Tgeo,Tparticle> &spc,

          Tenergy &pot,

          const Tpvec &p,

          const typename Space<Tgeo,Tparticle>::Change &c )

      {

          double du = 0;

          auto &g = spc.groupList();

          du += pot.g2All(p, c.mvGroup);

          du += pot.external(p);


          for ( auto &m : c.mvGroup ) // loop over all moved groups

          {

              size_t i = size_t(m.first);                       //
index of moved group


              du += pot.g_external(p, *g[i]);                   //
moved group <-> external


              if ( g[i]->isAtomic() )

              {                                                 //
Check if moved group is atomic

                  for ( auto j : m.second )                     //
loop over all moved groups

                      for ( auto k : *g[i] )                    //
loop over all atoms in moved group

                          if ( std::find(m.second.begin(),
m.second.end(), k) == m.second.end() || j

                              > k ) // check such that moved atoms
does not interact OR moved atoms interact only once

                              du += pot.i2i(p, j, k);

              }

              if ( g[i]->isMolecular() )

              {

                  if (!m.second.empty()) // only recalculate internal
energy if N>0

                     du += pot.g_internal(p, *g[i]);

              }

          }


          return du;

      }


NOTE: I changed the for ( auto j = i; j != mg.end(); j++ ) line to for (
auto j = (++mg.begin()); j != mg.end(); j++ )
Was it correct? Why calculate self interaction?

        virtual double g2All(const Tpvec & p, const std::map<int,
vector<int>>& mg)

        {

            double du = 0;

            auto &g = spc->groupList();


            for ( auto &m : mg ) // loop over all moved groups

            {

                size_t i = size_t(m.first);                       //
index of moved group


                // Calculate energy moved <-> static groups

                for ( size_t j = 0; j < g.size(); j++ ) {           //
loop over through all groups

                    if ( mg.count(j) == 0 ) {                // If
group j is not in mvGroup

                        du += g2g(p, *g[i], *g[j]);           // moved
group<->static groups

                        if ( du == pc::infty )

                            return pc::infty;   // early rejection

                    }

                }

            }


            // Calculate energy moved <-> moved

            for ( auto i = mg.begin(); i != mg.end(); i++ )

            {

                for ( auto j = (++mg.begin()); j != mg.end(); j++ ) //
MIKAEL should it not be j = (++mg.begin()) instead of j = i, why
include interaction with self

                {

                    auto _i = i->first;

                    auto _j = j->first;

                    du += g2g(p, *g[_i], *g[_j]);

                    if ( du == pc::infty )

                        return pc::infty;   // early rejection

                }

            }


            return du;

        }

LukasSukenik:
Actually this breaks the tests(polymer, membrane ) for some reason. Is this
the bug with internal energy you were talking about?

Anyway i will keep the g2All() function but revert the for cycles to the
original implementation

for ( auto i = mg.begin(); i != mg.end(); i++ )

            {

                for ( auto j = i; j != mg.end(); j++ )

mlund:
Hi Lukas!

About iterator loop, I do not think the self-interaction is included; it's just the way to loop over unique pairs using iterators. One more question about the groups: you use `molindex` to simply identify the group in the unordered map. Wouldn't it be possible to simply use the address, i.e. `&g`, as this will always be unique?

Best, Mikael

LukasSukenik:
Hello

Well, you could use the address, but you have a possible issue with
reallocation, which could happen during the simulation - although very
unlikely.

Should i change it to address rather than having a unique ID?

Best regards, Lukas

mlund:
Well, we already assume no group relocation occurs since the tracker operates with group pointers and also the group vector in Space contains pointers ;-)  If it works for EM and is just as fast, I'd prefer this approach to keep the complexity down. I'm drafting a more and robust and simpler design, but it will not be ready anytime soon.

LukasSukenik:
Hi

EnergyMatrix::GroupMap now uses addreses (&Group) as Keys.

As for using EnergyMatrix for atomic species - for example for deserno
membrane, It is much better to implement Verlet extended pairlist. I did
the benchmarks and it is 5 times faster with it than without for around
3000 lipids.

As for using float and half_floats it seems there isn't that much of a
difference in speed-up. I will do a more thorough benchmark, but from
preliminary results it seem that for 10 000 LJ particles the differences
are withing ~5% with the fastest being floats and slowest half_floats

A better strategy to reduce the load/update overhead as well as reducing
the memory requirements would be using cell list.

Best, Lukas

mlund:
Great, thanks! As a last thing, could you disable the playground in `src/CMakeLists.txt`? I want to remove the playground completely as it's not part of the Faunus library (one can still use `-DMYPLAYGROUND=...`). Too bad the half floats didn't give much speed-up; perhaps the floats are already in the cache. It think we should mention this somewhere in the paper.

LukasSukenik:
Ok, done. I commented out the playground references in src/CmakeLists.txt

I suppose we should mention the use of double data type in the article. The
question is whether type casting is making the float energy matrix slower.
I guess I will change all double to float in SC and run the speed-up
benchmark again.

PR #86, opened 2017-02-27
Updated dipole-particles, added 2D hypersphere, multipole-analysis fixed.

bjornstenqvist:


mlund:
Hej Björn, det ser ud til at der er konflikter. Måske du glemte at hente mine seneste (små) ændringer i nittests.*?

src/analysis.cpp
```diff
@@ -127,21 +127,6 @@ namespace Faunus
         f << energy() << "\n"; 
     }
 
-    void PairFunctionBase::normalize(data &d)
```
mlund: Why has this been removed from the base class and implemented separately in each derived class, i.e. `AtomRDF` and `MoleculeRDF`?

PR #110, opened 2019-02-01
Charge Fluctuations analysis

gitesei:


mlund:
Suggestion:
1. instead of specifying an atom index range w. `indexes`, specify a molecule id with `molecule`. In this way, multiple molecules would add to the average. Possible documentation:

`chargefluctuation`   |  description
---------- | ----------------- 
`molecule`            | Molecule name to analyse
`showall=False` | If `False` only fluctuating atoms are shown in output.

All molecules of the given name are analysed and a grand average is collected.

PR #112, opened 2019-02-05
updated charfluctuations analysis

mlund:

### DIFF
--- a/docs/_docs/analysis.md
+++ b/docs/_docs/analysis.md
@@ -228,14 +228,17 @@ The input keywords are:
 
 ## Charge Fluctuations
 
-Calculates the average charge, the standard deviation of the charge, and the most 
-probable species (atom name) of atomic indexes undergoing charge fluctuations.
+For a given molecule, this calculates the average charge, the standard deviation of the charge, 
+and the most probable species (atom name) averaging over all molecular groups.
+The structure of one of the molecular groups with average charges and most probable 
+atomic species is saved in PQR format.
 
 `chargefluctuations`    | Description
 ----------------------- | ----------------------------------------------------------
 `nstep`                 | Interval between samples.
 `nskip`                 | Number of initial steps excluded from the analysis.
-`indexes`               | Lower and upper bound of the index range to be analyzed.
+`molecule`              | Name of the molecule to analyze.
+`verbose`               | If True, include the results in the output file.
 
 ## Reaction Coordinate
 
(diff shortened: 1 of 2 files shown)


mlund:
Hi @gitesei, looks good. The code can be simplified by allowing for only _one_ molecule in the input, i.e. `molecules` -> `molecule`. If the user needs to analyse several molecules, it's merely to add another analysis instance in the input file.

src/analysis.h
```diff
@@ -736,58 +736,94 @@ namespace Faunus {
...
+                bool verbose = false;
+                size_t cnt;
 
                 void _sample() override {
-                    cnt = 0;
-                    for (auto p=spc.p.begin()+ndx0; p!=(spc.p.begin()+ndx1+1); ++p) {
-                        idcnt.at(cnt)[p->id]++;
-                        charge.at(cnt) += p->charge;
-                        ++cnt;
+                    for (auto &g : spc.groups) {
```
mlund: ```for (auto &g : spc.findMolecules(molid) {...}``` would simplify

src/analysis.h
```diff
@@ -736,58 +736,94 @@ namespace Faunus {
...
+                        cnt = 0;
+                        if (g.id == molid) {
+                            for ( auto p = g.begin(); p < g.trueend() ; ++p) {
```
mlund: Should this go to `trueend()`?

src/analysis.h
```diff
@@ -736,58 +736,94 @@ namespace Faunus {
...
+                    molname = j.at("molecule").get<decltype(molname)>(); // molecule name
+                    auto it = findName(Faunus::molecules<Tpvec>, molname);
+                    if (it == Faunus::molecules<Tpvec>.end())
+                        throw std::runtime_error("unknown species '" + molname + "'");
+                    molid = it->id(); // molecule ID
+                    idcnt.resize( it->atoms.size() );
+                    charge.resize( it->atoms.size() );
+                }
+
+                virtual ~ChargeFluctuations() {
```
mlund: cleaner to use the `FormatPQR` from io.h. I think it takes a vector of particles.

docs/_docs/analysis.md
```diff
@@ -228,14 +228,17 @@ The input keywords are:
 
 ## Charge Fluctuations
 
-Calculates the average charge, the standard deviation of the charge, and the most 
-probable species (atom name) of atomic indexes undergoing charge fluctuations.
+For a given molecule, this calculates the average charge, the standard deviation of the charge, 
+and the most probable species (atom name) averaging over all molecular groups.
+The structure of one of the molecular groups with average charges and most probable 
+atomic species is saved in PQR format.
```
mlund: Use instead a `pqrfile` keyword to let the user decide the name and if a pqr file should be saved?

src/analysis.h
gitesei: No, I agree. I got confused and here it should be auto &p : g. Later in the code, when I save the PQR file I should instead loop to the trueend() to make sure that the whole molecule is saved.

PR #114, opened 2019-02-12
stddev fix; infinite atom mass

mlund:


src/average.h
```diff
@@ -19,7 +19,7 @@ namespace Faunus
       } //!< Root-mean-square
 
       double stdev() const {
-          return std::sqrt( cnt / static_cast<double>(sqsum) - avg()*avg() );
+          return std::sqrt( (sqsum + avg()*avg() - 2*sum*avg()) / static_cast<double>(cnt-1) );
```
mlund: I think this should be:

`return std::sqrt( (sqsum + cnt*avg()*avg()- 2*sum*avg()) / static_cast<double>(cnt-1) )`

PR #122, opened 2019-04-05
Some updates

bjornstenqvist:


src/energy.h
```diff
@@ -1808,6 +1917,9 @@ namespace Faunus {
                                     if (it.key()=="confine")
                                         push_back<Energy::Confine<Tspace>>(it.value(), spc);
 
+                                    if (it.key()=="excludedvolume")
```
mlund: what is the difference between `Confine` and `ExcludedVolume` ?

bjornstenqvist: They are basically the inverse of one another. 'ExcludedVolume' is a 'hard' shape that is impregnable. But maybe it is better to still use 'Confine' and keywords like 'sphereInverse' or something. Yet, i.e. 'cylinder' can be a finite-length shape in 'ExcludedVolume' whereas that is more difficult to fix in 'Confine' with the used energy-function.

mlund: Could this be used instead: http://mlund.github.io/faunus/docs/energy/#custom-external-potential

bjornstenqvist: For s sphere or cuboid probably, but I do not think the cylinder.

bjornstenqvist: Or other more complex geometries which might be introduced.

mlund: Why would the external potential not work with a cylinder or other shapes? An arbitrary function can be given.

bjornstenqvist: It might, but it would probably get very ugly. And you would need the r-vector, not just r.norm().

mlund: No comprende :-) This example confines in a spherical shell. Why would confining outside a cylinder be less pretty? There's access to x,y,z coordinates of atoms and mass centra.

``` yaml
customexternal:
    molecules: [water]
    com: true
    constants: {radius: 15, dr: 3}
    function:
        var r2 := x^2 + y^2 + z^2;
        if ( r2 < radius^2 )
           1000 * ( radius-sqrt(r2) )^2;
        else if ( r2 > (radius+dr)^2 )
           1000 * ( radius+dr-sqrt(r2) )^2;
        else
           0;
```

mlund: Excluding a cylinder along z, one could use

``` yaml
customexternal:
    molecules: [water]
    com: true
    constants: {radius: 15}
    function:
        var r2 := x^2 + y^2;
        if ( r2 < radius^2 )
           1000 * ( radius-sqrt(r2) )^2; // or inf
        else
           0;
```

PR #127, opened 2019-04-24
Geometry

rc83:
A quick fix of linker's errors when using clang (forgotten inline keywords in cpp file). Override keyword used explicitly for overridden methods (clang warnings). As a bonus, unit tests of Chameleon verifying agreement between its own inline methods (vdist, boundary) and the underlying geometry implementations.

### DIFF
--- a/src/geometry.h
+++ b/src/geometry.h
@@ -642,6 +642,7 @@ namespace Faunus {
             }
 
             Chameleon(const Variant type = CUBOID);
+            Chameleon(const GeometryImplementation &geo, const Variant type);
 
             //! Copy everything, but clone the geometry.
             Chameleon(const Chameleon &geo) : GeometryBase(geo),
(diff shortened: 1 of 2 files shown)


src/geometry.h
```diff
@@ -663,214 +663,192 @@ namespace Faunus {
...
-                    }
-                } else {
-                    geometry->boundary(a);
+        inline void Chameleon::_setLength(const Point &l) {
+            len = l;
+            len_half = l * 0.5;
+            len_inv = l.cwiseInverse();
+        }
+
+        inline double Chameleon::getVolume(int dim) const {
```
mlund: getVolume(), setVolume(), randompos() etc. are not speed limiting so why suggest inlining?

rc83: As they are one-line wrappers only, hence to remove the possible (unimportant) overhead. Just a matter of habit. Can un-inline and move to cpp.

mlund: sure, it's a minor detail just to keep the exe size down and leave room for inlined functions of importance. I merge now and you can always revisit it later.

PR #130, opened 2019-04-26
functionparser compiles to a stand-alone object file

rc83:
Exprtk is a resources-hungry library during compilation, considerably
prolonging the compilation time and increasing the memory usage. Its
isolation to a standalone object file hugely improves compilation
characteristics, especially when  when the potentials are changed and
recompiled.

make -j2 faunus
touch src/potentials.h
time make -j2 faunus

before: system 182% cpu 2:27,81; peak memory usage approx 4.8 GB
after:  system 168% cpu 1:13,53; peak memory usage approx 2.5 GB
gcc 8.2.0; i5-4310U @ 2.00GHz


mlund:
Nice!

PR #163, opened 2019-07-05
Use yaml.safe_load() instead of deprecated yaml.load() in yason.py

rc83:
Tested only in Gentoo; tests in MacOSX before merge welcome.


rc83:
My mistake, sorry. I forget to restore the name when I tested the exceptions. Can you correct that or shall I?

PR #169, opened 2019-07-18
Refactor pair potentials to use PairMatrix computed by PairMixer

### LINKED ISSUE #159: Reimplement mixing rules for non-bonded potentials
1. The coefficient matrices are implemented like PairMatrix inside the respective potential.
2. PairMatrix is constructed using a Mixer class which takes a vector of AtomType, a lambda function to extract the property (e.g., sigma, epsilon) from the AtomType and an actual combining function (e.g., arithmetic or geometric average). We do not assume the mixing depends on other parameters, i.e., the sigma value can be computed solely from a pair of sigma values.
3. The Potential class can override any default pair parameters. That will be used for custom pair values. The custom pair values are internally treated as AtomType.
4. Serialization to JSON shall check if the parameters in the matrices correspond to the mixing rules; otherwise they shall be output as custom pairs.
5. Serialization from JSON shall be greatly simplified.

rc83:
The combination rules are separated from the pair potentials hence any
combination of them is theoretically possible. Creation of the potential
coefficient table is delegated to the PairMixer class. The new pair potential
as well as the combination rule can be added now effortlessly.
Close #159.


rc83:
I tried to write both the title and the message in a way that they both can be directly taken for the squash merge. ;-)

src/potentials.h
```diff
@@ -8,27 +8,138 @@
...
+#include "spdlog/sinks/null_sink.h"
 
 namespace Faunus {
 namespace Potential {
 
 using namespace std::string_literals;
 
+//! type of a matrix containing pair potential coefficients
+typedef Eigen::MatrixXd TPairMatrix;
+typedef std::shared_ptr<TPairMatrix> TPairMatrixPtr;
```
mlund: these typedefs seem to be connected with the mixer yet appear in the enclosing namespace. Why not confine them in `PairMixer`?

src/potentials.h
```diff
@@ -8,27 +8,138 @@
...
+void from_json(const json &j, std::vector<InteractionData> &interactions);
+void to_json(json &j, const std::vector<InteractionData> &interactions);
+
+/**
+ * @brief Known named combination rules for parameters of pair potential interaction.
+ *
+ * When adding a new one, add a json mapping. Also consider appending the PairMixer::getCombinator()
+ * method to recognize the new rule.
+ */
+enum ECombinationRule {COMB_UNDEFINED = 0, COMB_ARITHMETIC, COMB_GEOMETRIC, COMB_LORENTZ_BERTHELOT};
```
mlund: Perhaps `CombinationRuleType`?

src/potentials.h
```diff
@@ -8,27 +8,138 @@
...
+    {COMB_GEOMETRIC, "geometric"},
+    {COMB_LORENTZ_BERTHELOT, "lorentz_berthelot"},
+    {COMB_LORENTZ_BERTHELOT, "LB"}, // alternative non-canonical name
+})
+
+/**
+ * @brief Exception for handling pair potential initialization.
+ */
+struct PairPotentialException : public std::runtime_error {
+    PairPotentialException(const std::string msg) : std::runtime_error (msg) {};
```
mlund: add inline or move to .cpp

src/potentials.h
```diff
@@ -8,27 +8,138 @@
...
+    PairMixer(TExtractorFunc extractor, TCombinatorFunc combinator, TModifierFunc modifier = &modIdentity) :
+            extractor(extractor), combinator(combinator), modifier(modifier) {};
+
+    //! @return a square matrix of atoms.size()
+    TPairMatrixPtr createPairMatrix(const std::vector<AtomData> &atoms);
+    //! @return a square matrix of atoms.size()
+    TPairMatrixPtr createPairMatrix(const std::vector<AtomData> &atoms,
+                                    const std::vector<InteractionData> &interactions);
+
+    enum ECoefficient {COEF_ANY, COEF_SIGMA, COEF_EPSILON};
```
mlund: Skip the `E` prefix or spell out as `EnumCoefficient`

src/potentials.h
```diff
@@ -8,27 +8,138 @@
...
+    static TCombinatorFunc getCombinator(ECombinationRule combination_rule, ECoefficient = COEF_ANY);
+
+    // when explicit custom pairs are the only option
+    inline static double combUndefined(double, double) { return std::numeric_limits<double>::signaling_NaN(); };
```
mlund: add `constexpr`

src/potentials.h
mlund: `CoefficientType`?

src/potentials.h
```diff
@@ -8,27 +8,138 @@
...
+    inline static double combSum(double a, double b) { return (a+b); }
```
mlund: use `std::plus`

src/potentials.h
```diff
@@ -8,27 +8,138 @@
...
+};
 
 void to_json(json &j, const PairPotentialBase &base);   //!< Serialize any pair potential to json
 void from_json(const json &j, PairPotentialBase &base); //!< Serialize any pair potential from json
 
+/**
+ * @brief A common ancestor for potentials that use parameter matrices computed from atomic
+ * properties and/or custom atom pair properties.
+ */
+class MixerPairPotentialBase : public PairPotentialBase {
```
mlund: add virtual destructor

src/potentials.h
```diff
@@ -81,46 +190,62 @@ struct Dummy : public PairPotentialBase {
...
+                 ECombinationRule combination_rule = COMB_LORENTZ_BERTHELOT)
+        : MixerPairPotentialBase(name, cite, combination_rule) {};
 
     inline Point force(const Particle &a, const Particle &b, double r2, const Point &p) override {
-        double s6 = powi(m->s2(a.id, b.id), 3);
+        double s6 = powi((*sigma_squared)(a.id, b.id), 3);
         double r6 = r2 * r2 * r2;
         double r14 = r6 * r6 * r2;
-        return 6. * m->eps(a.id, b.id) * s6 * (2 * s6 - r6) / r14 * p;
+        return 6. * (*epsilon_quadrupled)(a.id, b.id) * s6 * (2 * s6 - r6) / r14 * p;
```
mlund: `epsilon_quadrupled` sounds to me like eps^4 (I know now that it's times four.) What about `four_times_epsilon`

src/potentials.h
```diff
@@ -251,22 +404,22 @@ struct RepulsionR3 : public PairPotentialBase {
...
+
   public:
-    Hertz(const std::string &name = "hertz");
+    Hertz(const std::string &name = "hertz")
+        : MixerPairPotentialBase(name) {};
     inline double operator()(const Particle &a, const Particle &b, const Point &r) const override {
         double r2 = r.squaredNorm();
-        if (r2 <= m->hd(a.id, b.id))
-            return m->ehe(a.id, b.id) * pow((1 - (sqrt(r2) / m->hd(a.id, b.id))), 2.5);
+        if (r2 <= (*hydrodynamic_diameter)(a.id, b.id))
```
mlund: I dislike `hydrodynamic`. What about just using `sigma` as for LJ, WCA, Hardsphere?

src/potentials.h
```diff
@@ -277,21 +430,21 @@ class Hertz : public PairPotentialBase {
...
+
   public:
-    SquareWell(const std::string &name = "squarewell");
+    SquareWell(const std::string &name = "squarewell")
+        : MixerPairPotentialBase(name) {};
     inline double operator()(const Particle &a, const Particle &b, const Point &r) const override {
-        if (r.squaredNorm() < m->th(a.id, b.id)) // threshold squared
-            return -m->esw(a.id, b.id);
+        if (r.squaredNorm() < (*diameter_sw_squared)(a.id, b.id))
+            return -(*depth_sw)(a.id, b.id);
```
mlund: `depth_sw` --> `epsilon_sw`?

src/potentials.h
rc83: Powers are based on geometric shapes, i.e., squared, cubed, while multiples are expressed as doubled, tripled, quadrupled, etc. I would keep it.

src/potentials.h
rc83: The resolution of the Hertz potential is already prepared in the next (queued) pull request.

src/potentials.h
rc83: The resolution of the Hertz potential is already prepared in the next (queued) pull request.

src/potentials.h
rc83: Done.

src/potentials.h
rc83: I removed the function completely as it is not used at all.

src/potentials.h
rc83: Done. My bad.

src/potentials.h
rc83: Constructors with an initializer list and empty body are used a lot in the header file. I can either move them all to cpp file or keep them all in the header file. The inline keyword is actually redundant as definition within the class makes the constructor inlined automatically. Suggestion?

src/potentials.h
rc83: Both the `PairMixer` class and descendants of `MixerPairPotentialBase` class share the equal interest in this data type. I would argue that this data type is meaningful even without the existence of the `PairMixer` class as the matrices can be produced in other ways.

src/potentials.h
rc83: Renamed.

src/potentials.h
rc83: Renamed to `PotentialCoefficient`.

rc83: Done.

PR #170, opened 2019-07-19
Correct the Hertz potential formula with a diameter squared

rc83:


src/potentials.h
```diff
@@ -396,29 +397,28 @@ struct RepulsionR3 : public PairPotentialBase {
 
 /**
  * @brief Hertz potential
- * @details This is a repulsive potential describes the change in elasticenergy of two deformable objects when subjected
- * to an axialcompression.
+ * @details This is a repulsive potential, that for example describes the change in elastic energy
+ * of two deformable objects when subjected to an axial compression.
```
mlund: rename R_h to sigma and remove any mention of "hydrodynamic"

src/potentials.cpp
```diff
@@ -731,18 +731,18 @@ void HardSphere::initPairMatrices() {
...
 void Hertz::initPairMatrices() {
-    const TExtractorFunc extract_radius = [](const AtomData &a) -> double { return a.hdr; };
+    const TExtractorFunc extract_diameter = [](const AtomData &a) -> double { return 2*a.hdr; };
     const TExtractorFunc extract_epsilon = [](const AtomData &a) -> double { return a.eps_hertz; };
-    const TCombinatorFunc comb_radius = PairMixer::getCombinator(combination_rule, PairMixer::COEF_SIGMA);
+    const TCombinatorFunc comb_diameter = PairMixer::getCombinator(combination_rule, PairMixer::COEF_SIGMA);
     const TCombinatorFunc comb_epsilon = PairMixer::getCombinator(combination_rule, PairMixer::COEF_EPSILON);
 
-    hydrodynamic_diameter = PairMixer(extract_radius, comb_radius, [](double x) -> double { return 2*x; })
+    diameter_squared = PairMixer(extract_diameter, comb_diameter, &PairMixer::modSquared)
```
mlund: `sigma_squared` like in LJ, WCA, hardsphere?

src/potentials.h
rc83: I thought about it and R_H could also mean a Hertz radius, cf. used eps_H. To stay consistent with `AtomData` I would keep the naming scheme as it is so far. However I already have a plan how to refactorize `AtomData` and pair potentials together with names changes.

src/potentials.cpp
rc83: I would keep the naming scheme as it is so far. However I already have a plan how to refactorize `AtomData` and pair potentials together with names changes.

src/potentials.h
mlund: OK, but I would like "hydrodynamic" to be eradicated. It should never have been there in the first place.

rc83: Kill it before it breed! :-D I amended the last commit and force-pushed. Not even a single mention about hydrodynamics. ;-)

Issue #160, opened 2019-07-03
Provide debug output to the user


rc83:
A lightweight logging framework shall be incorporated into faunus to enable efficient logging at various levels (e.g., error, warning, notice, debug). The logging framework shall be also capable to choose different sinks like stderr and/or file. The log level and the sink shall be command line options. Currently only messages to stderr are output.

Some examples of use cases: 1. When adding a new feature you realize that some combinations of parameters will not be currently used and hence tested thus a notice or warning shall be emitted. 2. An obsolete syntax in json configuration is encountered. 3. Very improbable state is detected.

Next step: Review logging framework candidates.

mlund:
Libraries for this exists:
- https://github.com/gabime/spdlog
- https://github.com/Emanem/elog
- https://github.com/PlatformLab/NanoLog (fastest, but requires c++17)

As a very simple solution, one could just push messages to a json array, possibly through an interface.

rc83:
The [spdlog](https://github.com/gabime/spdlog) looks neat and fits our needs. Important features: [Good documentation](https://github.com/gabime/spdlog/wiki), active development (on github), easy usage, reasonably fast, python-like string formatting.

I have added spdlog as an ExternalProject in `CMakeList.txt`. It compiles nicely and faunus cheers the world in colors for the purpose of testing. :-D

mlund:
Nice! The docs states that it's possible to link with an object file for faster compilation (instead of header-only). I don't think this feature is available in the last release, though. In any case, we should probably pin the version/commit. By "debug", I suppose you mean debugging usage, not code development? We should decide on a minimum set of levels, for example "info", "warn", "critical"?

rc83:
I have formulated the configuration to link the library as an object file thus I believe it is done so. ;-) Also the most recent commit of the library is explicitly pinned by its hash value in `CMakeList.txt`. However we should update the library revision regularly as its development is quite agile.

Yeah, by debug I meant debugging for the user, e.g., how the configuration options are understood, changes in the box volume, etc. A minimal reasonable set of levels could be:

error – a reason to terminate the program: invalid input, unrecoverable run condition, unable to use files
warn – the user shall rather change the simulation setup: suspicious input, recoverable run condition possibly impacting the results
debug – various status messages

Later we can split the levels as follows if needed: debug+info, notice+warn, error+critical. Levels warn and more sever should be output by default.

mlund:
A few things to condider:
1. simulations using MPI
[...]

PR #173, opened 2019-07-19
Use global instances of loggers (issue #160)

rc83:


rc83:
pyfaunus relies on default initialization with a null sink in the `core.cpp` file

PR #177, opened 2019-07-24
Add an aliasing of the mixing potentials coefficient names

### LINKED ISSUE #171: Dynamic atom type properties
The `AtomData` structure wires statically all possible properties of the atom type, though they may be used only at a single specific point, e.g., in a coefficient matrix of a pair potential. Moreover addition of a new pair potential often means extension of the `AtomType` structure.

The atomic properties which are (solely) pair potential coefficients should have a free structure, e.g., the keyword-float pairs. The potentials will then have additional arguments with reasonable defaults, mapping the potential coefficient to the atomic properties, e.g., sigma: "sigma", epsilon: "epsilon". The only danger I can see so far is a conflict, if we decide to use a name chosen by the user, as a keyword with a different purpose in a later version of Faunus.

This approach reduces the `AtomData` size as well as makes the potentials decoupled from the hardcoded atomic properties. An easy representation could be a dynamically allocated Eigen matrix with cols/rows identified by atom types and rows/cols by used coefficient types. The mapping of the coefficient names to the matrix indices could be stored as a vector or similarly. Refactorization of `AtomData` and potentials shall be quite straight forward. However, it shall be remembered that `AtomData` can be access from any place in the code.

Critical discussion welcome.

rc83:
To unify the user interface, the pair potentials coefficients are named as sigma and eps in the topology. However if multiple potentials are used at the same time, different value sets may be needed. Hence we allow to create an alias, e.g., sigma: sigma_hs, which overrides the eponymous default mapping. The naming of variables was adjusted accordingly.
    
AtomData from_json function is now scavenging like a vulture, eating anything looking like a number. If a builder for the state object was introduced, the potentials could announce their parameter sets in advance, enabling much cleaner way to get the topology parameters.
    
AtomData parameters, except sigma, are not access in any time-critical part of the code, hence simple string-double map is used. A copy of sigma is temporarily kept in an AtomData member.

Closes #171.


mlund:
thanks! pyfaunus compilation fails, see travis.

src/core.h
```diff
@@ -75,7 +75,7 @@ namespace Faunus {
      * keys as the object should be zero after being processed by
      * i.e. `from_json` or similar.
      */
-    struct SingleUseJSON : private json {
+    struct SingleUseJSON : public json {
```
mlund: This is potentially dangerous as you could access without deleting the entry which is promised by the class

src/potentials.cpp
```diff
@@ -134,19 +134,21 @@ void MixerPairPotentialBase::from_json(const json &j) {
...
         }
         if (j.count("custom") == 1) {
-            custom_pairs = j;
+            *custom_pairs = j;
         }
     } catch (const PairPotentialException &e) {
         faunus_logger->error(std::string(e.what()) + " in potential " + name);
         throw std::runtime_error("error deserialising potential " + name + " from json");
     }
+    extractorsFromJson(j);
```
mlund: a `//` comment would be nice here

docs/_docs/energy.md
```diff
@@ -303,19 +303,36 @@ while zero for $r>r\_c+w\_c$.
...
+topology.
 
-type           | $\epsilon$       | $u(r)$
--------------- | ---------------- | --------------------------------------------------------
-`hardsphere`   |                  | $\infty$ for $r < \sigma\_{ij}$
-`hertz`        | `eps_hertz`      | $\epsilon\_{ij} \left ( 1-r / \sigma\_{ij}\right )^{5/2}$ for $r<\sigma\_{ij}$
-`lennardjones` | `eps`            | $4\epsilon\_{ij} \left ( (\sigma\_{ij}/r\_{ij})^{12} - (\sigma\_{ij}/r\_{ij})^6\right )$
-`squarewell`   | `eps_sw`         | $-\epsilon\_{ij}$ for $r<\sigma\_{ij}$ 
-[`wca`](http://dx.doi.org/ct4kh9) | `eps` | $u\_{ij}^{\text{LJ}} + \epsilon\_{ij}$ for $r < 2^{1/6}\sigma\_{ij}$
+type           | atomic parameters | $u(r)$ (non-zero part)
```
mlund: please check that `make manual` passes (pdf creation)

rc83: Errors corrected in the force-push. Just for convenience Charter instead of Garamond font was used.

src/core.h
rc83: I am aware of that. I believe a stricter validation of json will be introduced anyway with the builder and creator patterns.

src/potentials.cpp
rc83: I have added a comment locally, which I will commit later as a part of source code house-keeping.

PR #185, opened 2019-09-04
Move examples in separate directories

rc83:
Examples can produce files with identical names. This is troublesome,
especially when running them in parallel.


rc83:
Unless it would ease automation in some way I would rather keep the prefixes as a matter of convenience. Personally I do not like to have many files with the same name but different paths unless it is a sort of database. It is a source of mistakes by overlooking.

rc83:
Unless it would ease automation in some way I would rather keep the prefixes as a matter of convenience. Personally I do not like to have many files with the same name but different paths unless it is a sort of database. It is a source of mistakes by overlooking.

PR #189, opened 2019-09-16
Freesasa

rc83:
Use FreeSASA instead of PowerSASA

### DIFF
--- a/docs/_docs/energy.md
+++ b/docs/_docs/energy.md
@@ -662,7 +662,7 @@ elements of `high`.
 ## Solvent Accessible Surface Area
 
 Note that the implementation of Solvent Accessible Surface Area potential is considered _experimental_.
-The code is untested, unoptimized and the configuration syntax bellow can change.
+The code is untested, unoptimized and the configuration syntax below can change.
 The FreeSASA library option has to be enabled when [compiling].
 
 `sasa`       | SASA Transfer Free Energy
--- a/docs/_docs/energy.md
+++ b/docs/_docs/energy.md
@@ -357,7 +357,7 @@ as shown in the example bellow.
       Na Cl: {sigma_hs: 2}
 ~~~
 
-### SASA
+### SASA (pair potential)
 
 This calculates the surface area of two intersecting particles or radii $R$ and $r$
 to estimate an energy based on transfer-free-energies (TFE) and surface tension.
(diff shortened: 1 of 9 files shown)


docs/_docs/energy.md
```diff
@@ -660,6 +661,10 @@ elements of `high`.
 
 ## Solvent Accessible Surface Area
 
+Note that the implementation of Solvent Accessible Surface Area potential is considered _experimental_.
+The code is untested, unoptimized and the configuration syntax bellow can change.
```
mlund: below

src/energy.cpp
```diff
@@ -508,34 +508,86 @@ void Hamiltonian::sync(Energybase *basePtr, Change &change) {
...
-void SASAEnergy::updateSASA(const ParticleVector &p) {
-    assert(ps != nullptr);
+
+#ifdef ENABLE_FREESASA
+
+SASAEnergy::SASAEnergy(Space &spc, double cosolute_concentration, double probe_radius)
+    : spc(spc), cosolute_concentration(cosolute_concentration)
+{
+    name = "sasa"; // todo predecessor constructor
+    cite = "doi:10.12688/f1000research.7931.1"; // todo predecessor constructor
```
mlund: I try to use the short doi service which gives "doi:10/dbjh"

src/energy.cpp
```diff
@@ -508,34 +508,86 @@ void Hamiltonian::sync(Energybase *basePtr, Change &change) {
...
+    parameters = freesasa_default_parameters;
+    parameters.probe_radius = probe_radius * 1.0_angstrom;
+    init();
+}
+
+SASAEnergy::SASAEnergy(const json &j, Space &spc) : spc(spc) {
```
mlund: call the first constructor from the second constructor (allowed from C++11)

src/energy.cpp
```diff
@@ -508,34 +508,86 @@ void Hamiltonian::sync(Energybase *basePtr, Change &change) {
         }
     throw std::runtime_error("hamiltonian mismatch");
 }
-void SASAEnergy::updateSASA(const ParticleVector &p) {
-    assert(ps != nullptr);
+
+#ifdef ENABLE_FREESASA
+
+SASAEnergy::SASAEnergy(Space &spc, double cosolute_concentration, double probe_radius)
```
mlund: doxygen comment w. @param stating expected units

PR #209, opened 2019-10-22
Fix Gromos torsion potential

rc83:
Fix Gromos torsion potential and add extensive unit tests for bonds. For details see the individual commit messages.

Suitable for **Rebase and merge** merging strategy.


rc83:
I believe the codacity issue is not justified.

src/bonds.cpp
```diff
@@ -67,15 +67,18 @@ void BondData::shift(int offset) {
 
 bool BondData::hasEnergyFunction() const { return energy != nullptr; }
 
-BondData::~BondData() {}
+BondData::BondData(const decltype(index) index) : index(index) {}
```
mlund: pass by reference

src/bonds.cpp
```diff
@@ -67,15 +67,18 @@ void BondData::shift(int offset) {
...
+
+HarmonicBond::HarmonicBond(double k, double req, const decltype(index) index)
```
mlund: by ref &

src/bonds.cpp
```diff
@@ -153,51 +156,59 @@ void FENEWCABond::setEnergyFunction(const ParticleVector &p) {
...
     aeq = j.at("aeq").get<double>() * 1.0_deg;
 }
 
 void HarmonicTorsion::to_json(Faunus::json &j) const {
-    j = {{"k", k / (1.0_kJmol / std::pow(1.0_rad, 2))}, {"aeq", aeq / 1.0_deg}};
+    j = {{"k", 2 * k_half / (1.0_kJmol / std::pow(1.0_rad, 2))}, {"aeq", aeq / 1.0_deg}};
     _roundjson(j, 6);
 }
 
+HarmonicTorsion::HarmonicTorsion(double k, double aeq, const decltype(index) index)
```
mlund: ref &

src/bonds.cpp
mlund: `decltype` is cryptic and for understanding the signature, I think explicitly using `vector<int>` is better. I'd reserve `decltype()` for more complex types or when using templates. Another alternative is a typedef, but that's also an extra layer.

src/bonds.h
```diff
@@ -33,21 +33,25 @@ struct BondData {
...
+    double k_half = 0, req = 0;
     int numindex() const override;
     Variant type() const override;
     std::shared_ptr<BondData> clone() const override;
     void from_json(const json &j) override;
     void to_json(json &j) const override;
     std::string name() const override;
     void setEnergyFunction(const ParticleVector &p);
+    HarmonicBond() = default;
+    HarmonicBond(double k, double req, decltype(index) index);
```
mlund: ref &

src/bonds.cpp
```diff
@@ -153,51 +156,59 @@ void FENEWCABond::setEnergyFunction(const ParticleVector &p) {
...
+    cos_aeq = cos(j.at("aeq").get<double>() * 1.0_deg); // cos(angle)
 }
 
 void GromosTorsion::to_json(Faunus::json &j) const {
-    j = {{"k", 2.0 * k / 1.0_kJmol}, {"aeq", std::acos(aeq) / 1.0_deg}};
+    j = {{"k", 2 * k_half / 1.0_kJmol}, {"aeq", std::acos(cos_aeq) / 1.0_deg}};
 }
 
-BondData::Variant GromosTorsion::type() const { return BondData::G96_TORSION; }
+GromosTorsion::GromosTorsion(double k, double cos_aeq, const decltype(index) index)
```
mlund: &

src/bonds.cpp
```diff
@@ -206,18 +217,22 @@ int PeriodicDihedral::numindex() const { return 4; }
...
+    n = j.at("n").get<double>();               // multiplicity/periodicity n
+    phi = j.at("phi").get<double>() * 1.0_deg; // angle
 }
 
 void PeriodicDihedral::to_json(Faunus::json &j) const {
-    j = {{"k", k[0] / 1.0_kJmol}, {"n", k[1]}, {"phi", k[2] / 1.0_deg}};
+    j = {{"k", k / 1.0_kJmol}, {"n", n}, {"phi", phi / 1.0_deg}};
 }
 
+PeriodicDihedral::PeriodicDihedral(double k, double phi, double n, const decltype(index) index)
```
mlund: &

rc83:
Changed to passing the vector by reference and the explicit `std::vector<int>` type. I think that we should introduce type aliases by `typedef` for commonly used indices like particle index or atom index and use them consistently.

mlund:
`vector<int>` is a pretty basic type, somewhat close to an old-style c-array. Introducing typedefs adds an extra layer of complexity and the caller would have to acquire knowledge of the underlying type (could it be a list, vector, valarray, set, or something else?). A template function could handle them all, but of course at the cost of compile time and potential code bloat.

PR #212, opened 2019-10-29
Bondvector

rc83:
# Description
Refactor `BondData` vector to use `BasePointerVector`

- The excluded_neighbour generator now takes into account any 1-2 bond, not only harmonic ones.
- Add unit tests for FENE and FENEWCA bonds.

Includes also PR #211 to allow subsequent merging.

### DIFF
--- a/src/bonds.h
+++ b/src/bonds.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "core.h"
+#include "auxiliary.h"
 #include "particle.h"
 
 namespace Faunus {
@@ -154,13 +155,15 @@ struct PeriodicDihedral : public BondData {
  * Serialize to/from json
  */
 
-void to_json(json &j, const std::shared_ptr<BondData> &b);
 void from_json(const json &j, std::shared_ptr<BondData> &b);
+void to_json(json &j, const std::shared_ptr<const BondData> &b);
+void to_json(Faunus::json &j, const BondData &b);
 
 void setBondEnergyFunction(std::shared_ptr<BondData> &b,
                            const ParticleVector &p); //!< Set the bond energy function of `BondData` which
                                                      //!< require a reference to the particle vector
 
+[[deprecated("Use bonds.find<TClass>() method instead.")]]
 inline auto filterBonds(const std::vector<std::shared_ptr<BondData>> &bonds, BondData::Variant bondtype) {
     std::vector<std::shared_ptr<BondData>> filt;
     filt.reserve(bonds.size());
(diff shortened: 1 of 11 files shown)


src/molecule.h
```diff
@@ -53,64 +66,147 @@ struct Conformation {
...
+ * @internal MoleculeData uses this class internally.
+ */
+class ExclusionsVicinity {
+    //! count of atoms in the molecule; unmutable
+    int atoms_cnt;
+    //! maximal possible distance (difference) between indices of excluded particles
+    int  max_bond_distance;
+    //! 1D representation of excluded interaction matrix; the shared pointer saves copying
+    //! Note that std::vector<bool> uses packing with performance implications
+    std::shared_ptr<std::vector<unsigned char>> excluded_pairs;
```
mlund: what is `unsigned char` storing? Big enough?

src/bonds.h
```diff
@@ -119,19 +149,21 @@ struct PeriodicDihedral : public BondData {
...
+void to_json(json &j, const std::shared_ptr<const BondData> &b);
+void to_json(Faunus::json &j, const BondData &b);
 
 void setBondEnergyFunction(std::shared_ptr<BondData> &b,
                            const ParticleVector &p); //!< Set the bond energy function of `BondData` which
                                                      //!< require a reference to the particle vector
 
+[[deprecated("Use bonds.find<TClass>() method instead.")]]
 inline auto filterBonds(const std::vector<std::shared_ptr<BondData>> &bonds, BondData::Variant bondtype) {
     std::vector<std::shared_ptr<BondData>> filt;
```
mlund: I forgot why BondData needs to be stored as pointers? This could lead to excessive memory fragmentation compared to a straight vector.

rc83: `BondData` is an abstract class for all bonded interactions hence pointers. In general I do not worry so much if the data structure is accessed only outside the MC loop.

src/molecule.cpp
```diff
@@ -320,11 +591,11 @@ ParticleVector &Conformation::toParticleVector(ParticleVector &p) const {
     assert(not p.empty() and not empty());
     // copy positions
     if (positions.size() == p.size())
-        for (size_t i = 0; i < p.size(); i++)
+        for (int i = 0; i < p.size(); i++)
             p[i].pos = positions[i];
     // copy charges
     if (charges.size() == p.size())
-        for (size_t i = 0; i < p.size(); i++)
+        for (int i = 0; i < p.size(); i++)
```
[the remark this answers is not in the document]
rc83: This piece has been already merged in PR #211. It is introduced by amistake. I will prepare a separate fix in another PR.

Issue #214, opened 2019-11-03
Performance issue with newer versions of Faunus


mlund:
Below is a benchmark of the `examples/bulk` system of melted NaCl for the same number particles, trial moves, release mode; and all analysis disabled. Hardcoded, non-splined potentials are used in both cases. 

1. faunus-v1.0.2:  **54 secs**
2. faunus-master (f5e11ef10107d11a441b2c1dd08906cf860fc2bd): **105 secs**

Possible places to look:
- [x] `geometry::vdist()`
- [ ] passing of `change`
- [x] compare with nonbonded switched off, i.e. probe only trial move code
- [ ] compare clang and gcc

ping @rc83

mlund:
Letting `Potential::Coulomb` and `WeeksChandlerAndersen` return constant 1.0, only reduces the time to 80 secs, indicating that the bottleneck is not the pair-potential. Completely skipping `g_internal()` in Nonbonded, the simulation finishes in a few seconds. Next target, `vdist()`.

rc83:
My guess based on some quick tests: The difference between 75 s and 105 s is at least partly caused by (not) inlining of the `vdist` function. When I accidentally kept some garbage in a branch of `vdist` which was not executed at all, the run time increased from 75 s to 90 s in my case.

mlund:
On my system, omitting the first `ORTHOGONAL` if-stement reduces time quite to ~75 secs which could be the threshold for the compiler to use inlining. Still, I'm surprised that the the raw `a-b` distance is so much quicker, i.e. 35 secx. I'd like to try with GCC, but I currently cannot make it link on the mac setup (brew or cleanup may be needed, hrmf).

rc83:
Clang inlines the `Chameleon::vdist` function when the following patch is applied. I have a three seconds difference between hardcoded conditions for orthogonal coordinates and PBC and their runtime evaluation. However there is still a 15 seconds penalty for evaluating the PBC.
 
```diff --git a/src/geometry.h b/src/geometry.h
--- a/src/geometry.h
+++ b/src/geometry.h
@@ -397,10 +397,6 @@ inline Point Chameleon::vdist(const Point &a, const Point &b) const {
             else if (distance.z() < -len_half.z())
                 distance.z() += len.z();
         }
-    } else if (boundary_conditions.coordinates == ORTHOHEXAGONAL ||
-               boundary_conditions.coordinates == TRUNC_OCTAHEDRAL) {
-        distance = a - b;
[...]

PR #231, opened 2019-12-25
Add support for Euclidian norm in Nonbonded

mlund:
This commit gives a 20-40% speed improvement by
optimizing distance calculations; g_internal summation;
and by using aggressive optimization in Release mode.
Doctest and pybind11 are updated to latest releases.


mlund:
Closes #214

PR #238, opened 2020-01-15
Compare against theoretical values in the chain example

rc83:
Compare against theoretical values in the chain example


rc83:
This is a fix which shall come to version 2.3.

PR #241, opened 2020-01-22
Implement IPBC to scattering analysis

mlund:


mlund:
@rc83 add parallel version to scatter branch

PR #243, opened 2020-02-04
Scattering OMP parallelisation

rc83:
# Description

Refactor and paralellise scattering analysis
    
OpenMP is enabled for loops in the sample methods that have O(N^2) complexity, where N is number of particles.
    
Furthermore, the sample methods are rewritten to allow automatic vectorisation of trigonometric functions in GCC by using `libmvec` vector math library when `--ffast-math` is enabled. This speeds up the analysis by factor of 4 on modern processors.

## Checklist

- No unit tests were provided.
- [x] code naming scheme follows the convention
  - Constants for OMP ranges are single capital letters here, which is a grey zone of naming rules. Other rules should be honoured. Variables have been extensively renamed from a single letter physicist convention to English words to follow our programmers convention.
- [x] the source code is well documented
- [x] the user-manual is updated regarding the effective parallelisation
- [ ] performance is checked in supported configurations: (GCC, Clang) × (GNU/Linux, MacOSX)

## Note

The pull request includes and thus obsoletes the pull request #241.

### DIFF
--- a/src/analysis.cpp
+++ b/src/analysis.cpp
@@ -1339,6 +1339,7 @@ void ScatteringFunction::_to_json(json &j) const {
     case EXPLICIT:
         j["scheme"] = "explicit";
         j["pmax"] = explicit_average->pmax;
+        j["ipbc"] = explicit_average->ipbc;
         break;
     }
 }
@@ -1362,7 +1363,8 @@ ScatteringFunction::ScatteringFunction(const json &j, Space &spc) try : spc(spc)
     } else if (scheme_str == "explicit") {
         scheme = EXPLICIT;
         int pmax = j.value("pmax", 15);
-        explicit_average = std::make_shared<Scatter::StructureFactor<double>>(pmax);
+        bool ipbc = j.value("ipbc", false);
+        explicit_average = std::make_shared<Scatter::StructureFactor<double>>(pmax, ipbc);
         // todo: add warning if used a non-cubic system
     } else
         throw std::runtime_error("unknown scheme");
--- a/src/analysis.cpp
+++ b/src/analysis.cpp
@@ -1313,33 +1313,49 @@ void ScatteringFunction::_sample() {
     for (int id : ids) { // loop over molecule names
         auto groups = spc.findMolecules(id);
         for (auto &g : groups) // loop over groups
-            if (usecom && !g.atomic)
+            if (use_com && !g.atomic)
                 p.push_back(g.cm);
             else
                 for (auto &i : g) // loop over particle index in group
                     p.push_back(i.pos);
     }
+
+    // zero-padded suffix to use with `save_after_sample`
+    std::string suffix = fmt::format("{:07d}", cnt);
     switch (scheme) {
     case DEBYE:
         debye->sample(p, spc.geo.getVolume());
+        if (save_after_sample)
+            IO::write(filename + "." + suffix, debye->getIntensity());
         break;
-    case EXPLICIT:
-        explicit_average->sample(p, spc.geo.getLength().x());
+    case EXPLICIT_PBC:
+        explicit_average_pbc->sample(p, spc.geo.getLength().x());
+        if (save_after_sample)
+            IO::write(filename + "." + suffix, explicit_average_pbc->getSampling());
+        break;
+    case EXPLICIT_IPBC:
+        explicit_average_ipbc->sample(p, spc.geo.getLength().x());
+        if (save_after_sample)
+            IO::write(filename + "." + suffix, explicit_average_ipbc->getSampling());
         break;
     }
 }
 void ScatteringFunction::_to_json(json &j) const {
-    j = {{"molecules", names}, {"com", usecom}};
+    j = {{"molecules", names}, {"com", use_com}};
     switch (scheme) {
     case DEBYE:
         j["scheme"] = "debye";
-        j["qmin"] = debye->qmin;
-        j["qmax"] = debye->qmax;
+        std::tie(j["qmin"], j["qmax"], std::ignore) = debye->getQMeshParameters();
         break;
-    case EXPLICIT:
+    case EXPLICIT_PBC:
         j["scheme"] = "explicit";
-        j["pmax"] = explicit_average->pmax;
-        j["ipbc"] = explicit_average->ipbc;
+        j["pmax"] = explicit_average_pbc->getQMultiplier();
+        j["ipbc"] = false;
+        break;
+    case EXPLICIT_IPBC:
+        j["scheme"] = "explicit";
+        j["pmax"] = explicit_average_ipbc->getQMultiplier();
+        j["ipbc"] = true;
         break;
     }
 }
@@ -1347,7 +1363,8 @@ void ScatteringFunction::_to_json(json &j) const {
 ScatteringFunction::ScatteringFunction(const json &j, Space &spc) try : spc(spc) {
     from_json(j);
     name = "scatter";
-    usecom = j.value("com", true);
+    use_com = j.value("com", true);
+    save_after_sample = j.value("stepsave", false);   // save to disk for each sample
     filename = j.at("file").get<std::string>();
     names = j.at("molecules").get<decltype(names)>(); // molecule names
     ids = names2ids(molecules, names);                // names --> molids
@@ -1361,10 +1378,15 @@ ScatteringFunction::ScatteringFunction(const json &j, Space &spc) try : spc(spc)
         //    faunus_logger->warn("{0}: '{1}' scheme used on PBC system; consider 'explicit' instead", name,
         //    scheme_str);
     } else if (scheme_str == "explicit") {
-        scheme = EXPLICIT;
-        int pmax = j.value("pmax", 15);
         bool ipbc = j.value("ipbc", false);
-        explicit_average = std::make_shared<Scatter::StructureFactor<double>>(pmax, ipbc);
+        int pmax = j.value("pmax", 15);
+        if (ipbc) {
+            scheme = EXPLICIT_IPBC;
+            explicit_average_ipbc = std::make_shared<Scatter::StructureFactorIPBC<>>(pmax);
+        } else {
+            scheme = EXPLICIT_PBC;
+            explicit_average_pbc = std::make_shared<Scatter::StructureFactorPBC<>>(pmax);
+        }
         // todo: add warning if used a non-cubic system
     } else
         throw std::runtime_error("unknown scheme");
@@ -1375,10 +1397,13 @@ ScatteringFunction::ScatteringFunction(const json &j, Space &spc) try : spc(spc)
 void ScatteringFunction::_to_disk() {
     switch (scheme) {
     case DEBYE:
-        debye->save(filename);
+        IO::write(filename, debye->getIntensity());
         break;
-    case EXPLICIT:
-        explicit_average->save(filename);
+    case EXPLICIT_PBC:
+        IO::write(filename, explicit_average_pbc->getSampling());
+        break;
+    case EXPLICIT_IPBC:
+        IO::write(filename, explicit_average_ipbc->getSampling());
         break;
     }
 }
--- a/src/io.h
+++ b/src/io.h
@@ -27,6 +27,37 @@ bool writeFile(const std::string &file, const std::string &s,
 
 void strip(std::vector<std::string> &string_vector, const std::string &pattern); //!< Strip lines matching a pattern
 
+/**
+ * Write a map to an output stream as key-value pairs
+ * @tparam TKey
+ * @tparam TValue
+ * @param stream Output stream
+ * @param data
+ */
+template <typename TKey, typename TValue>
+void write(std::ostream &stream, const std::map<TKey, TValue> &data, const std::string &sep = " ", const std::string &end = "\n") {
+    if (stream) {
+        for (auto [key, value] : data) {
+            stream << key << sep << value << end;
+        }
+    }
+}
+
+/**
+ * Write a map to a file as key-value pairs
+ * @tparam TKey
+ * @tparam TValue
+ * @param filename
+ * @param data
+ */
+template<typename TKey, typename TValue>
+void write(const std::string &filename, const std::map<TKey, TValue>& data) {
+    if (!data.empty()) {
+        std::ofstream file(filename);
+        write(file, data);
+    }
+}
+
 } // namespace IO
 
 /**
--- a/src/io.h
+++ b/src/io.h
@@ -50,8 +50,7 @@ void write(std::ostream &stream, const std::map<TKey, TValue> &data, const std::
  * @param filename
  * @param data
  */
-template<typename TKey, typename TValue>
-void write(const std::string &filename, const std::map<TKey, TValue>& data) {
+template <typename TKey, typename TValue> void write(const std::string &filename, const std::map<TKey, TValue> &data) {
     if (!data.empty()) {
         std::ofstream file(filename);
         write(file, data);
--- a/src/scatter.h
+++ b/src/scatter.h
@@ -2,10 +2,12 @@
 
 #include <fstream>
 #include <algorithm>
+#include <cmath>
 
 namespace Faunus {
 
-/** @brief Routines related to scattering */
+/** @brief Routines related to scattering.
+ */
 namespace Scatter {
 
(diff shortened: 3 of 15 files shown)


src/scatter.h
```diff
@@ -33,222 +35,335 @@ template <class T = float> class FormFactorSphere {
...
+};
+
+
+/**
+ * A policy for collecting samples. To be used together with StructureFactor class templates.
+ *
+ * @tparam T float or double
+ */
+template <typename T> class SamplingPolicy {
+  public:
```
mlund: if paranoid, add `static_assert(std::is_floating_point<T>::value, "T must be floating point");`

src/scatter.h
```diff
@@ -33,222 +35,335 @@ template <class T = float> class FormFactorSphere {
...
+    };
+    typedef std::map<T, sampled_value> TSampledValues;
+  private:
+    TSampledValues samples;
+    const T precision = 10000.0; //!< precision of the key for better binning
+
+  public:
+    std::map<double, double> getSampling() {
+        std::map<double, double> average;
+        for (auto &kv : samples) {
```
mlund: `auto [key, value]`

src/scatter.h
```diff
@@ -33,222 +35,335 @@ template <class T = float> class FormFactorSphere {
...
+        const T key = std::round(key_approx * precision) / precision; // round |q| for better binning
+        samples[key].value += value;
+        samples[key].weight += weight;
     }
 };
 
+
 /**
- * @brief Structor factor calculation using explicit q averaging
+ * @brief Structure factor calculation using explicit q averaging.
```
mlund: remove punctuation

src/scatter.h
```diff
@@ -33,222 +35,335 @@ template <class T = float> class FormFactorSphere {
...
-     * @param groupList Vector of group pointers - i.e. as returned from `Space::groupList()`.
-     * @note Particle form factors are always set to unity, i.e. F(q) is ignored.
-     *
-     * @warning Untested
+     * @return a tuple of min, max, and step parameters of a q-mash
      */
-    template <class Tpvec, class Tg> void sampleg2g(const Tpvec &p, Tg groupList) {
-        assert(qmin > 0 && qmax > 0 && dq > 0 && "q range invalid.");
-        sampleg2g(p, qmin, qmax, dq, groupList);
+    std::tuple<T, T, T> getQMeshParameters() {
```
mlund: `auto`?

src/scatter.h
```diff
@@ -33,222 +35,335 @@ template <class T = float> class FormFactorSphere {
...
-                    }
-        int cnt = 0, N = p.size();
-        for (T q = qmin; q <= qmax; q += dq)
-            I[q] += 2. * _I.at(cnt++) / N + 1; // add to average I(q)
+    /**
+     * @return a map containing q (key) and average intensity (value)
+     */
+    std::map<double, double> getIntensity() {
+        std::map<double, double> averaged_intensity;
+        for (int m = 0; m < intensity.size(); ++m) {
```
mlund: int -> size_t

src/scatter.h
```diff
@@ -33,222 +35,335 @@ template <class T = float> class FormFactorSphere {
...
+        }
     }
 
-    void save(const std::string &filename) {
-        if (not S.empty()) {
-            std::ofstream f(filename.c_str());
-            if (f)
-                for (auto &i : S)
-                    f << i.first << " " << i.second / W[i.first] << "\n";
+    int getQMultiplier() {
```
mlund: is this for a future purpose?

src/scatter.h
```diff
@@ -33,222 +35,335 @@ template <class T = float> class FormFactorSphere {
...
+    using TSamplingPolicy::addSampling;
+
+  public:
+    StructureFactorIPBC(int q_multiplier) : p_max(q_multiplier){}
+
+    template <class Tpvec> void sample(const Tpvec &positions, const double boxlength) {
+        // https://gcc.gnu.org/gcc-9/porting_to.html#ompdatasharing
+        // #pragma omp parallel for collapse(2) default(none) shared(directions, p_max, positions, boxlength)
+        #pragma omp parallel for collapse(2) default(shared)
+        for (int i = 0; i < directions.size(); ++i) {
```
mlund: int -> size_t

src/scatter.h
```diff
@@ -33,222 +35,335 @@ template <class T = float> class FormFactorSphere {
...
+ *
+ * @tparam T float or double
+ */
+template <typename T> class SamplingPolicy {
+  public:
+    struct sampled_value {
+        T value;
+        T weight;
+    };
+    typedef std::map<T, sampled_value> TSampledValues;
```
mlund: "TSampledValues" does to reflect that it's a map containing both keys and values.

src/analysis.cpp
```diff
@@ -1311,40 +1311,60 @@ void ScatteringFunction::_sample() {
     for (int id : ids) { // loop over molecule names
         auto groups = spc.findMolecules(id);
         for (auto &g : groups) // loop over groups
-            if (usecom && !g.atomic)
+            if (use_com && !g.atomic)
```
mlund: Add one more check that the group is active for use with GCMC, typically by requiring that `g.size()==g.capacity()`

src/analysis.cpp
```diff
@@ -1311,40 +1311,60 @@ void ScatteringFunction::_sample() {
...
         break;
     }
 }
 
 ScatteringFunction::ScatteringFunction(const json &j, Space &spc) try : spc(spc) {
     from_json(j);
     name = "scatter";
-    usecom = j.value("com", true);
+    use_com = j.value("com", true);
+    save_after_sample = j.value("stepsave", false); // save to disk for each sample
```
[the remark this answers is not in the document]
mlund: And to `_docs/analysis.md`

src/scatter.h
```diff
@@ -33,222 +35,335 @@ template <class T = float> class FormFactorSphere {
...
+        if (q_step <= 0 || q_min <= 0 || q_max <= 0 || q_min > q_max ||
+            q_step / q_max < 4 * std::numeric_limits<T>::epsilon()) {
+            throw std::range_error("DebyeFormula: Invalid mesh parameters for q");
+        }
+        q_mesh_min = q_min < 1e-6 ? q_step : q_min; // ensure that q > 0
+        q_mesh_max = q_max;
+        q_mesh_step = q_step;
+        // resolution of the 1D mesh approximation of the scattering vector magnitude q
+        const size_t q_resolution = 1 + (size_t)((q_max - q_min) / q_step);
+        if (q_resolution > std::numeric_limits<int>::max()) { // we use int index variable
```
mlund: if overstepping, unsigned int becomes zero. can this condition evaluate to true?

src/analysis.cpp
```diff
@@ -1311,40 +1311,60 @@ void ScatteringFunction::_sample() {
...
 }
 void ScatteringFunction::_to_json(json &j) const {
-    j = {{"molecules", names}, {"com", usecom}};
+    j = {{"molecules", names}, {"com", use_com}};
     switch (scheme) {
     case DEBYE:
         j["scheme"] = "debye";
-        j["qmin"] = debye->qmin;
-        j["qmax"] = debye->qmax;
+        std::tie(j["qmin"], j["qmax"], std::ignore) = debye->getQMeshParameters();
```
mlund: In c++17, this should be valid: `auto [j["qmin"], j["qmax"], std::ignore) = debye->getQMeshParameters();]`

src/analysis.cpp
mlund: Actually, `findMolecules()` already by default set's `ACTIVE`, so never mind.

src/analysis.cpp
```diff
@@ -1311,40 +1311,60 @@ void ScatteringFunction::_sample() {
...
+            if (use_com && !g.atomic)
                 p.push_back(g.cm);
             else
                 for (auto &i : g) // loop over particle index in group
                     p.push_back(i.pos);
     }
+
+    // zero-padded suffix to use with `save_after_sample`
+    std::string suffix = std::to_string(cnt);
+    suffix = std::string(7 - suffix.length(), '0') + suffix;
```
mlund: here `fmt::format` would probably shine!

src/scatter.h
```diff
@@ -33,222 +35,335 @@ template <class T = float> class FormFactorSphere {
...
+                std::transform(positions.begin(), positions.end(), qr_products.begin(),
+                               [&q](auto &r) { return q.dot(r); });
+                // as of January 2020 the std::transform_reduce is not implemented in libc++
+                T sum_sin = 0;
+                for (auto &qr : qr_products) {
+                    sum_sin += std::sin(qr);
+                }
+                // as of January 2020 the std::transform_reduce is not implemented in libc++
+                T sum_cos = 0;
+                for (auto &qr : qr_products) {
```
mlund: How well does Eigen fare for this operation? It's supposed to handle vectorisation on a number of architectures. I believe the operation here would by `qr_products.sin().sum()`. Note also that for the coming Eigen 4, iterators are supported.

mlund: Btw. your sin/cos optimization may be applicable also in `energy.h:127` for Ewald summation.

mlund: Update: on GCC9/MacOS, the separate cos/sin (ifdef _GNUC_) are slower

mlund: The following gives a factor four speed-up on GCC and Clang:
~~~ cpp
Eigen::VectorXd qr_products;
qr_products.noalias() = positions * q;
T sum_sin = qr_products.array().sin().sum();
T sum_cos = qr_products.array().cos().sum();
~~~
where `positions` must by a `MatrixXd(N,3)` object. Best to await a unit-test to see if it's doing the right thing, though :-)

src/io.h
```diff
@@ -26,6 +26,26 @@ bool writeFile(const std::string &file, const std::string &s,
...
 
+/**
+ * Write a map to a file as key-value pairs
+ * @tparam TKey
+ * @tparam TValue
+ * @param filename
+ * @param data
+ */
+template<typename TKey, typename TValue>
+void write(const std::string &filename, const std::map<TKey, TValue>& data) {
```
mlund: Make an overload `write(std::ostring &, const std::map<> &)` to enable streaming to memory or compressed streams. Useful for unit testing. Perhaps rename to `writeMap`?

mlund: sorry `std::ostream`

src/scatter.h
```diff
@@ -33,222 +35,335 @@ template <class T = float> class FormFactorSphere {
...
+    T q_mesh_min, q_mesh_max, q_mesh_step; //<! q_mesh parameters in inverse angstrom; used for inline lambda-functions
 
-        if (dq <= 0 || qmin <= 0 || qmax <= 0 || qmin > qmax)
-            throw std::runtime_error("DebyeFormula: invalid q parameters");
+    /**
+     * @param m mesh point index
+     * @return the scattering vector magnitude q at the mesh point m
+     */
+    #pragma omp declare simd uniform(this) linear(m:1)
+    T q_mesh(int m) {
```
mlund: `inline`

src/io.h
```diff
@@ -26,6 +26,26 @@ bool writeFile(const std::string &file, const std::string &s,
...
+    if (!data.empty()) {
+        std::ofstream f(filename.c_str());
```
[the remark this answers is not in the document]
rc83: Done. This code was just copy-pasted.

src/io.h
```diff
@@ -26,6 +26,26 @@ bool writeFile(const std::string &file, const std::string &s,
...
+        if (f) {
+            for (auto &d : data) {
```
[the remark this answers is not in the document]
rc83: Done.

src/scatter.h
rc83: Done.

src/scatter.h
rc83: It is used in `analysis.cpp` in `to_json` function.

src/analysis.cpp
rc83: The structure binding does not like array-like identifiers in the list. Kept.

src/scatter.h
rc83: Done, though the function is already `inline` implicitly.
> A function defined entirely inside a class/struct/union definition, whether it's a member function or a non-member friend function, is implicitly an inline function.

src/scatter.h
```diff
@@ -33,222 +35,335 @@ template <class T = float> class FormFactorSphere {
...
+    static constexpr T r_cutoff_infty = 1e9; //<! a cutoff distance in angstrom considered to be infinity
+    T q_mesh_min, q_mesh_max, q_mesh_step; //<! q_mesh parameters in inverse angstrom; used for inline lambda-functions
 
-        if (dq <= 0 || qmin <= 0 || qmax <= 0 || qmin > qmax)
-            throw std::runtime_error("DebyeFormula: invalid q parameters");
+    /**
+     * @param m mesh point index
+     * @return the scattering vector magnitude q at the mesh point m
+     */
+    #pragma omp declare simd uniform(this) linear(m:1)
```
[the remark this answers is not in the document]
rc83: It is not an architecture-specific directive. It just hints the compile about how this SIMD-enabled function should be treated in SIMD vectorization. As the function is very simple, the compiler may know anyway.
[Documentation](https://www.openmp.org/spec-html/5.0/openmpsu42.html#x65-1530002.9.3.3)

src/analysis.cpp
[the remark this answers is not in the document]
rc83: How come that this could come like this from the pull request #241? :-D Done. Documentation shall also contain `ipbc`. As `ipbc` has no meaning for `debye` scheme, shall we keep it as an independent attribute, or shall we have three different schemes?

src/scatter.h
```diff
@@ -33,222 +35,335 @@ template <class T = float> class FormFactorSphere {
...
+        T value;
+        T weight;
+    };
+    typedef std::map<T, sampled_value> TSampledValues;
+  private:
+    TSampledValues samples;
+    const T precision = 10000.0; //!< precision of the key for better binning
+
+  public:
+    std::map<double, double> getSampling() {
```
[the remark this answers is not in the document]
rc83: I meant it on purpose to have accumulators always in at least `double` precision. As only `float` and `double` will be probably used for `T` hard coding seems reasonable for me. I have added a comment.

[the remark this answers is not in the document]
rc83: Done.

src/scatter.h
rc83: It could theoretically work also with ints. Kept.

src/scatter.h
rc83: changed to a proper English sentence instead

src/scatter.h
rc83: Done. Let’s be modern! ;-)

src/scatter.h
rc83: Let’s do it properly. Inspired by boost, I add `numeric_cast` function into `auxiliary.h`. Here I cast and throw an exception.

src/scatter.h
rc83: Done.

src/scatter.h
rc83: Done.

src/analysis.cpp
rc83: Done.

src/io.h
rc83: It is not needed. To keep pragmatically the code base small, I won’t fix it

src/scatter.h
rc83: Any suggestion on the name? (This is the last point to be resolved.)

mlund: TSampledValueMap? Analogous we have TParticleVector in Space.

rc83: Done.

rc83:
1. I had to revert one commit originating from the `master` branch to make the pull request compile.
2. There is no code related to performance testing yet; neither in `scatter.h` nor `scatter_test.h`. I will prepare another commit tomorrow.

**Create a merge commit** is probably the best strategy how to merge this pull request.

src/analysis.cpp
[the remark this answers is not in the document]
mlund: yes, debye is quite different and also requres dq, qmin, qmx while these are automatic w. explicit and ipbc that om the other hand require qmax. I would keep ipbc as an option in the explicit scheme and split debye to a separate analysis.

PR #255, opened 2020-02-19
Minor update in manual, and added charge compensation in Ewald

bjornstenqvist:
# Description

Please edit this text to include a summary of the change and which issue is fixed.

## Checklist

- [ ] `make test` passes with no errors
- [ ] the source code is well documented
- [ ] new functionality includes unittests
- [ ] the user-manual has been updated to reflect the changes
- [ ] I have installed the provided commit hook to auto-format commits (requires `clang-format`):

  ``` bash
  ./scripts/git-pre-commit-format install
  ```

- [ ] code naming scheme follows the convention:

  Style        | Elements
  ------------ | ---------------------
  `PascalCase` | classes, namespaces
  `camelCase`  | functions
  `snake_case` | variables


src/energy.cpp
```diff
@@ -291,24 +291,31 @@ double PolicyIonIon::surfaceEnergy(const EwaldData &d, Change &change, Space::Tg
 }
 
 double PolicyIonIon::selfEnergy(const EwaldData &d, Change &change, Space::Tgvec &groups) {
```
mlund: `PolicyIonIon::selfEnergy()` is never used as the self-energy is injected by the pair-potential which in this case means from the external CoulombGalore.

src/clustermove.cpp
```diff
@@ -7,7 +7,7 @@ namespace Move {
 
 double Cluster::clusterProbability(const Cluster::Tgroup &g1, const Cluster::Tgroup &g2) const {
     if (spc.geo.sqdist(g1.cm, g2.cm) <= group_thresholds(g1.id, g2.id))
-        return 1.0;
+        return probability;
```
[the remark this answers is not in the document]
mlund: Also, this should probably be in a separate PR as it seems somewhat unrelated to the topic of Ewald(?)

docs/_docs/energy.md
```diff
@@ -202,45 +202,48 @@ If `shift=false`, the potential is left unshifted and any given cutoff is ignore
...
 ### Ewald Summation
 
-If type is `ewald`, terms from reciprocal space; surface energies; and
-self energies are automatically added to the Hamiltonian, activating additional keywords:
+If type is `ewald`, terms from reciprocal space and surface energies are automatically added (in addition to the previously mentioned self- and real space-energy) to the Hamiltonian which activates the additional keywords:
 
 `type=ewald`          | Description
 --------------------- | ---------------------------------------------------------------------
-`kcutoff`             | Reciprocal-space cutoff
+`ncutoff`             | Reciprocal-space cutoff (unitless)
```
mlund: The keyword used in the code is `kcutoff`

docs/_docs/energy.md
```diff
@@ -202,45 +202,48 @@ If `shift=false`, the potential is left unshifted and any given cutoff is ignore
...
 `epss=0`              | Dielectric constant of surroundings, $\varepsilon_{surf}$ (0=tinfoil)
 `ewaldscheme=PBC`     | Periodic (`PBC`) or isotropic periodic ([`IPBC`](http://doi.org/css8)) boundary conditions
 `spherical_sum=true`  | Spherical/ellipsoidal summation in reciprocal space; cubic if `false`.
```
mlund: `spherical_sum` is missing from `docs/schema.yml`; could you add it so that the input is properly checked?

mlund: (just search for Ewald in the file and you'll see where to insert it)

src/clustermove.cpp
```diff
@@ -43,7 +43,7 @@ void Cluster::_to_json(json &j) const {
     }
 }
 void Cluster::_from_json(const json &j) {
-    assertKeys(j, {"dp", "dprot", "dir", "threshold", "molecules", "repeat", "satellites", "dirrot"});
+    assertKeys(j, {"dp", "dprot", "dir", "threshold", "molecules", "repeat", "satellites", "spread", "dirrot"});
```
mlund: You can delete this line completely as it's already covered by `docs/schema.yml`

rc83:
@mlund please let me know when the PR is ready for merge. Seeing the current PR history, I will do it cleanly. Except few extravagant commit messages, we are now keeping the git history tidy for release notes.

docs/_docs/energy.md
bjornstenqvist: I have changed it to ncutoff (which it should be) everywhere I could find.

src/energy.cpp
```diff
@@ -9,7 +9,7 @@ namespace Energy {
 EwaldData::EwaldData(const json &j) {
     alpha = j.at("alpha");                          // damping-parameter
     r_cutoff = j.at("cutoff");                      // real space cut-off
-    k_cutoff = j.at("kcutoff");                     // reciprocal space cut-off
```
mlund: This breaks backwards compatibility, so please use something like this:
~~~ cpp
if (j.count("kcutoff") {
    faunus_logger->warn("`kcutoff` is deprecated, use `ncutoff` instead");
    n_cutoff = j.at("kcutoff");
} else {
    n_cutoff = j.at("ncutoff");
}
~~~
Also, please keep the `kcutoff` parameter in the schema for now. Why is `ncutoff` btw. better than `kcutoff`?

bjornstenqvist: n is a vector of integers, k=2*pi*n/L a vector of real numbers. Those two often get mixed up.

mlund: ahh, ok. Should't then `n_cutoff` be an `int`? (it's currently a `double` and also in the schema it should probably be `integer` instead of `number`)

bjornstenqvist: Each element in the n-vector is an integer. However, as the length of a general n-vector is a double, for example n=(2,1,0) gives the double sqrt(5), the cutoff for the n-vector should also be allowed to be a double. Or no?

mlund: I think it's ok with a double then; in the code it's rounded with `ceil` anyways

rc83:
And I made a mistake. Please do not pull.

PR #259, opened 2020-02-28
Allow implicit molecules in speciation move

mlund:
This patch implements implicit molecules to be used in chemical reactions involving finite reservoirs of solid compounds.

- [x] `Space` is responsible for keeping track of implicit molecules.
  - State handled by json I/O.
  - Data is hidden and accessed via get-function
- [x] User input `insertmolecules` used also for implicit molecules; strengthened input check for `insertMolecules()`
- [x] Unused `ReactionData::canonic` and `ReactionData::N_reservoir` has been abandoned
- [x] Implicit molecules can occur as both reactants and products and stoichiometric coefficients may be given
- [x] Implicit molecules can be used in multiple coupled reactions
- [x] `Move::Speciation` tracks average reservoir sizes
- [x] A new test-case for calcium hydroxide precipitation is included. Output verified against code by Christophe Labbez.
- [x] Above example is implemented both using atomic groups and molecular groups. The latter is significantly slower, possibly due to `findMolecules()`, see #260

Squash commit with edited comments intended.


mlund:
- Codacy warning is a false positive; ignored.
- Failing macos MPI build seems to be a travis bug(?)

PR #263, opened 2020-03-08
Add "molarity" keyword for insertmolecules

mlund:
When inserting molecules, "molarity" can now be used as an alternative to "N". This commit also refactors the `Faunus::insertMolecules` into a separate class with where responsibilities are split among several (hidden) functions.

### DIFF
--- a/src/space.cpp
+++ b/src/space.cpp
@@ -359,9 +359,10 @@ bool InsertMoleculesInSpace::insertAtomicGroups(MoleculeData &moldata, Space &sp
     }
     return true;
 }
-void InsertMoleculesInSpace::insertMolecularGroups(MoleculeData &moldata, Space &spc, int N, bool inactive) {
+void InsertMoleculesInSpace::insertMolecularGroups(MoleculeData &moldata, Space &spc, int num_molecules,
+                                                   bool inactive) {
     assert(moldata.atomic == false);
-    while (N-- > 0) { // insert molecules
+    while (num_molecules-- > 0) { // insert molecules
         spc.push_back(moldata.id(), moldata.getRandomConformation(spc.geo, spc.p));
         if (inactive) {
             spc.groups.back().unwrap(spc.geo.getDistanceFunc());
@@ -369,10 +370,11 @@ void InsertMoleculesInSpace::insertMolecularGroups(MoleculeData &moldata, Space 
         }
     }
 }
-bool InsertMoleculesInSpace::setPositionsForTrailingGroups(int N, Space &spc, const Faunus::ParticleVector &positions,
+bool InsertMoleculesInSpace::setPositionsForTrailingGroups(Space &spc, int num_molecules,
+                                                           const Faunus::ParticleVector &positions,
                                                            const Point &offset) {
-    assert(spc.groups.size() >= N);
-    if (positions.size() == N * (spc.groups.end() - N)->traits().atoms.size()) {
+    assert(spc.groups.size() >= num_molecules);
+    if (positions.size() == num_molecules * (spc.groups.end() - num_molecules)->traits().atoms.size()) {
         size_t j = spc.p.size() - positions.size();
         for (auto i : positions) {
             i.pos = i.pos + offset;
@@ -384,7 +386,7 @@ bool InsertMoleculesInSpace::setPositionsForTrailingGroups(int N, Space &spc, co
             }
         }
         if (j == positions.size()) {
-            for (auto g = spc.groups.end() - N; g != spc.groups.end(); ++g) {
+            for (auto g = spc.groups.end() - num_molecules; g != spc.groups.end(); ++g) {
                 g->cm = Geometry::massCenter(g->begin(), g->end(), spc.geo.getBoundaryFunc(), -g->begin()->pos);
             }
         }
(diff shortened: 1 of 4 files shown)


src/space.cpp
```diff
@@ -435,9 +340,128 @@ void makeNaCl(Space &space, int num_particles, const Geometry::Chameleon &geomet
...
+            spc.groups.back().unwrap(spc.geo.getDistanceFunc());
+            spc.groups.back().resize(0);
+        }
+    }
+}
+bool InsertMoleculesInSpace::setPositionsForTrailingGroups(Space &spc, int num_molecules,
+                                                           const Faunus::ParticleVector &particles,
+                                                           const Point &offset) {
+    assert(spc.groups.size() >= num_molecules);
+    if (particles.size() == num_molecules * (spc.groups.end() - num_molecules)->traits().atoms.size()) {
```
rc83, line 377: I believe that it is silently assumed here that all trailing groups have the same number of atoms, e.g., are of the same molecule type. A comment would be handy here what is the purpose of the condition.

src/space.cpp
```diff
@@ -435,9 +340,128 @@ void makeNaCl(Space &space, int num_particles, const Geometry::Chameleon &geomet
...
+        size_t cnt = spc.p.size() - particles.size();
```
rc83, line 378: A more descriptive variable name please. It is more like a particle index than count. After changing the loop it will be a offset of the particle vector.

src/space.cpp
```diff
@@ -435,9 +340,128 @@ void makeNaCl(Space &space, int num_particles, const Geometry::Chameleon &geomet
...
+        for (auto &i : particles) {
+            auto pos = i.pos + offset;
+            if (spc.geo.collision(pos) == false) {
+                spc.p.at(cnt++).pos = pos;
+            } else {
+                faunus_logger->warn("position outside box");
+                return false;
+            }
+        }
+        if (cnt == particles.size()) {
```
rc83, line 388: When is this condition not true? What is its purpose?

src/space.cpp
```diff
@@ -435,9 +340,128 @@ void makeNaCl(Space &space, int num_particles, const Geometry::Chameleon &geomet
...
+        size_t cnt = spc.p.size() - particles.size();
+        for (auto &i : particles) {
+            auto pos = i.pos + offset;
+            if (spc.geo.collision(pos) == false) {
+                spc.p.at(cnt++).pos = pos;
+            } else {
+                faunus_logger->warn("position outside box");
+                return false;
+            }
+        }
```
rc83, line 387: When iterating simultaneously over two structures, I consider more readable to use single index in a for-loop and two offsets, here one for `particles` and the other for `spc.p`.

src/space.cpp
```diff
@@ -435,9 +340,128 @@ void makeNaCl(Space &space, int num_particles, const Geometry::Chameleon &geomet
...
+                    if (concentration > pc::epsilon_dbl) {
+                        double rel_error = (concentration - num_molecules / spc.geo.getVolume()) / concentration;
+                        if (rel_error > 0.01) {
+                            faunus_logger->warn("{}: initial molarity differs by {}% from target value", molname,
+                                                rel_error * 100);
+                        }
+                    }
+                }
+                if (not moldata->isImplicit() and num_molecules < 1) {
+                    throw ConfigurationError(molname + ": at least one molecule required");
```
rc83, line 430: I would add a probable cause to the error message: concentration too low.

src/space.cpp
```diff
@@ -435,9 +340,128 @@ void makeNaCl(Space &space, int num_particles, const Geometry::Chameleon &geomet
...
+                    double concentration = properties.at("molarity").get<double>() * 1.0_molar;
+                    num_molecules = std::round(concentration * spc.geo.getVolume());
+                    if (concentration > pc::epsilon_dbl) {
+                        double rel_error = (concentration - num_molecules / spc.geo.getVolume()) / concentration;
+                        if (rel_error > 0.01) {
+                            faunus_logger->warn("{}: initial molarity differs by {}% from target value", molname,
+                                                rel_error * 100);
+                        }
+                    }
+                }
```
rc83, line 428: A function returning `num_molecules` and `rel_error` would be nice. The remaining code here would just log a warning if `rel_error` is to big.

src/space.cpp
```diff
@@ -435,9 +340,128 @@ void makeNaCl(Space &space, int num_particles, const Geometry::Chameleon &geomet
...
+                if (not moldata->isImplicit() and num_molecules < 1) {
+                    throw ConfigurationError(molname + ": at least one molecule required");
+                }
+                bool inactive = properties.value("inactive", false); // active or not?
+                {
+                    std::string state = (inactive) ? "inactive" : "active";
+                    if (moldata->isImplicit()) {
+                        state = "implicit";
+                    }
```
rc83, line 437: readability: `if(implicit) {…} else if (inactive) {…} else {…}`

src/space.cpp
mlund: now using iterator instead

src/space.cpp
mlund: fixed

src/space.cpp
mlund: next time

src/space.cpp
mlund: fixed

mlund:
Thanks for the through review; mostly everything fixed

PR #266, opened 2020-03-19
Test for multipoledistribution analysis

mlund:
This adds a test case for the multipole distribution analysis, `multipoledist`, and also fixes a problem with the way the quadrupole moment was calculated, giving wrong results for the distribution analysis. Squash merge intended.


examples/multipole/multipole.yml
```diff
@@ -0,0 +1,47 @@
...
+            - neg2: [-0.5, 0, 0]
+            - pos3: [0.5, 0, 0]
+    - quadrupole:
+        rigid: true
+        insdir: [0,0,1]
+        structure:
+            - pos1: [1.0, 1.0, 0.0]
+            - pos1: [-1.0, -1.0, 0.0]
+            - neg1: [1.0, -1.0, 0.0]
+            - neg2: [-1.0, 1.0, 0.0]
```
bjornstenqvist: should it not be neg1?

mlund: It's on purpose since I want all moments to be non-zero in the test

bjornstenqvist: Good then!

PR #271, opened 2020-04-23
Update install.md

ohyh:
To turn on MPI, isn't `-DENABLE_MPI=ON` correct rather than old `-DENABLE_MPI=OFF`?

### DIFF
--- a/docs/_docs/install.md
+++ b/docs/_docs/install.md
@@ -70,7 +70,7 @@ The following options are available:
 
 CMake Option                         | Description
 ------------------------------------ | ---------------------------------------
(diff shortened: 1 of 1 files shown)


mlund:
The `=OFF` indicates the default value if nothing is specified. Perhaps modify the text to clarify that? 
Note that MPI is used only in special cases (parallel tempering and penalty functions) and will generally not speed things up at the moment.

mlund:
Do you want to update the PR or should I close this?

ohyh:
> Do you want to update the PR or should I close this?

It is OK for me to close it. Thanks!

PR #277, opened 2020-05-02
Draw from active molecular groups

mlund:
For molecular groups draw only from active ones while for atomic groups, select ALL. This fixes a potential problem for grand canonical, molecular groups subject to internal changes using `AtomicTranslateRotate` is used. In the old code, the random molecule would be drawn from a distribution that contained also inactive molecules while it should include only active ones.


mlund:
@gitesei: this could affect LNP lipids, albeit likely a minor bias

Issue #283, opened 2020-06-09
Force calculation for bonded interactions


mlund:
Add `force()` function for `Energy::Bonded` interactions for use with Langevin dynamics (#282).

- [x] Harmonic bonds
- [x] Bonds involving angles
- [x] Unittests

Allen and Tildesley's Appendix C could be useful and they refer to (f90) example code on their [Github rep](/Allen-Tildesley) for angular bond forces. For C++ code, perhaps [OpenMM's implementation](/openmm/openmm/blob/master/platforms/reference/src/ReferenceKernels.cpp) may be useful.

PR #295, opened 2020-06-26
Forces for bonded interactions for use with Langevin dynamics

mlund:
This adds force calculations to bonded interactions and, with harmonic as an example, shows how forces can be implemented for arbitrary bond types. These implementations should/could be done in a separate PR.

- [x] add `force` functor to `BondData`
- [x] add `Bonded::force()`
- [x] implement force calculation in `Harmonic` incl. unittest
- [ ] double-check correctness of `Harmonic::forceFunc`. Sign on force OK?

Contributes to #283

### DIFF
--- a/src/energy.cpp
+++ b/src/energy.cpp
@@ -610,10 +610,11 @@ void Bonded::force(std::vector<Point> &forces) {
             assert(bond->forceFunc != nullptr);                    // the force function must be implemented
             auto bond_forces = bond->forceFunc(distance_function); // get forces on each atom in bond
             assert(bond->index.size() == bond_forces.size());
+            int j = 0;
             for (int index : bond->index) { // loop over atom index in bond (relative to group begin)
                 auto absolute_index = std::distance(spc.p.begin(), group.begin()) + index;
                 assert(absolute_index < forces.size());
-                forces[absolute_index] += bond_forces[index]; // add to overall force
+                forces[absolute_index] += bond_forces[j++]; // add to overall force
             }
         }
     }
--- a/src/energy.cpp
+++ b/src/energy.cpp
@@ -599,13 +599,15 @@ double Bonded::energy(Change &change) {
  * - loop over inter-molecular bonds and _add_ force
  *
  * Force unit: kT/Å
+ *
+ * @warning Untested
  */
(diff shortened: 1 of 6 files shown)


src/energy.cpp
```diff
@@ -589,6 +589,47 @@ double Bonded::energy(Change &change) {
...
+        for (auto bond : bonds) {                                  // loop over all bonds in group
+            assert(bond->forceFunc != nullptr);                    // the force function must be implemented
+            auto bond_forces = bond->forceFunc(distance_function); // get forces on each atom in bond
+            assert(bond->index.size() == bond_forces.size());
+            int j = 0;
+            for (int index : bond->index) { // loop over atom index in bond (relative to group begin)
+                auto absolute_index = std::distance(spc.p.begin(), group.begin()) + index;
+                assert(absolute_index < forces.size());
+                forces[absolute_index] += bond_forces[j++]; // add to overall force
+            }
```
rc83, line 618: If `bond->index` is a vector, I would prefer to use a single loop index variable `i` to address both `bond->index[i]` and `bond_forces[i]`. The non-loop variable `j` is confusing.

src/energy.cpp
```diff
@@ -589,6 +589,47 @@ double Bonded::energy(Change &change) {
...
+    }
+
+    for (auto bond : inter) {                              // loop over inter-molecular bonds
+        assert(bond->forceFunc != nullptr);                // the force function must be implemented
+        auto bond_forces = bond->forceFunc(distance_function); // get forces on each atom in bond
+        assert(bond_forces.size() == bond->index.size());
+        int j = 0;
+        for (int absolute_index : bond->index) {        // loop over atom index in bond (absolute index)
+            forces[absolute_index] += bond_forces[j++]; // add to overall force
+        }
```
rc83, line 629: If `bond->index` is a vector, I would prefer to use a single loop index variable `i` to address both `bond->index[i]` and `bond_forces[i]`. The non-loop variable `j` is confusing.

src/energy.cpp
mlund: fixed

src/energy.cpp
mlund: fixed

Issue #282, opened 2020-06-09
Langevin dynamics implementation


SHervoe-Hansen:
The implementation of Langevin dynamics in Faunus for a Lennard-Jones fluid/gas is well underway in my fork of <a href="https://github.com/SHervoe-Hansen/faunus/tree/forcemove_ld">Faunus</a>. Checklist which should preferably be handled as separate PRs:
- [x] Working example of LJ system
- [ ] Unit testing
  - simple LJ system. RDF, temperature, avg. kinetic and potential energy? Compare w. MC
  - harmonic oscillator?
- [ ] Force calculation implementation:
  - #284: Reciprocal ewald space
  - #283: Bonded interactions (harmonic, torsion, dihedrals -- see [OpenMM implementation](https://github.com/openmm/openmm/blob/master/platforms/reference/src/ReferenceKernels.cpp)) and also Allen+Tildesley, Appendix C
  - `FunctorPotential` for combined potentials
  - Anisotropic non-bonded interactions?
  - See Allen and Tildesley's Appendix C as well as their [Github rep](/Allen-Tildesley) with examples of force calculations (ewald, angles, dipoles etc.)
- [ ] Error handling
  - [ ] Gracefully exit from explosion (closes #305)
- [ ] Documentation (`moves.md`)
- [ ] An initialisation scheme for simulation of liquids (slow heating, minimisation using 0 K Monte Carlo).
- [ ] Improved control over MD/MC switching.
- [ ] Parallel scheme for force calculation in `Nonbonded` class (OMP pragmas?)

himalayanpride:
Is it possible for me to get the beta version of the code where you have implemented langevin dynamics? I want to test CG martini system ( I don't have any angle/dihedral terms in the system). 
Thanks,

PR #298, opened 2020-07-02
Add LangevinDynamics move

rc83:
# Description

Add LangevinDynamics move including user documentation and a minimal example of a LJ fluid.

Thorough review is welcome as (i) the functionality is very new and (ii) the branch split 6 months ago hence several conflicts had to be addressed.

Other related but different fixes are also provided in separate commits, thus ‘rebase and merge’ is a good strategy for merging.

Contributes to #282.

## Checklist

- [ ] `make test` passes with no errors
- [ ] the source code is well documented
- [ ] new functionality includes unittests
- [ ] the user-manual has been updated to reflect the changes
- [ ] I have installed the provided commit hook to auto-format commits (requires `clang-format`):


src/forcemove.h
```diff
@@ -0,0 +1,98 @@
...
+#include "energy.h"
+#include "units.h"
+
+namespace Faunus {
+
+namespace Move {
+
+/**
+ * @brief Base class for dynamics integrators
+ *
```
mlund, line 12: Add a little bit more info about the responsibilities

src/forcemove.h
```diff
@@ -0,0 +1,98 @@
...
+ */
+class IntegratorBase {
+  protected:
+    Space &spc;
+    Energy::Energybase &energy;
+    IntegratorBase(Space &spc, Energy::Energybase &energy);
```
mlund, line 18: remove variable names

src/forcemove.h
```diff
@@ -0,0 +1,98 @@
...
+    virtual ~IntegratorBase() = default;
+
+  public:
+    /**
+     * @brief Move particles by one time step and update their positions, velocities and acting forces
+     * @param velocities
```
mlund, line 24: Only keep brief doxygen comment in header, the rest in cpp file

src/forcemove.h
```diff
@@ -0,0 +1,98 @@
...
+void from_json(const json &j, IntegratorBase &i);
+void to_json(json &j, const IntegratorBase &i);
+
+/**
+ * @brief Symmetric Langevin velocity-Verlet method (BAOAB)
+ *
+ * The integrator can conduct normal velocity-Verlet, when gamma = 0.
+ */
+class LangevinVelocityVerlet : public IntegratorBase {
+    float dt;    //!< integration time step (picoseconds)
```
mlund, line 41: - why float? everything else in faunus is double... 
- rename `dt` to `time_step` at least as the internal variable name

src/forcemove.h
```diff
@@ -0,0 +1,98 @@
...
+    float gamma; //!< friction coefficient (inverse picoseconds)
+
+    //! increment particle's position in an A semi-step
+    inline Point positionIncrement(const Point& velocity);
```
mlund, line 45: put doxygen comment _after declaration, i.e. `; //!< increment particle's position in an A semi-step`. Remove variable name

src/forcemove.h
```diff
@@ -0,0 +1,98 @@
...
+ *
+ * Orchestrate execution of integrators, thermostats, etc. Store vectors for velocities and forces, which are not part
+ * of the particle vector in the space.
+ */
+class ForceMoveBase : public Movebase {
+  protected:
+    unsigned int nsteps; //!< number of integration steps to perform during a single MC move
+    Space &spc;          //!< space with particles and their positions
+    std::shared_ptr<IntegratorBase> integrator;
+    PointVector velocities, forces; //!< velocities and forces; the vector index corresponds to the particle index
```
mlund, line 72: Add a way to load/save velocities from disk? It could be a function that takes a json formatted stream and simply looks for the keys `velocities` and, possibly, `forces`. We could then either use a separate file, or read from / append to the existing `state.json` file.

src/forcemove.h
```diff
@@ -0,0 +1,98 @@
...
+  public:
+    double bias(Change &, double, double) override;
+};
+
+/**
+ * @brief Langevin dynamics move using Langevin equation of motion
+ */
+class LangevinDynamics : public ForceMoveBase {
+  public:
+    LangevinDynamics(Space &spc, std::shared_ptr<IntegratorBase> integrator, unsigned int nsteps = 0);
```
mlund, line 90: - Is there a point in giving `nsteps` a default value?
- Remove variable names

src/forcemove_test.h
```diff
@@ -0,0 +1,43 @@
...
+#include "externalpotential.h"
+
+namespace Faunus {
+namespace Move {
+
+using doctest::Approx;
+
+TEST_SUITE_BEGIN("ForceMove");
+
+TEST_CASE("[Faunus] Integerator") {
```
mlund, line 12: Integrator

src/move.h
```diff
@@ -6,6 +6,7 @@
 #include "geometry.h"
 #include "auxiliary.h"
 #include "space.h"
+#include "energy.h"
```
mlund, line 9: use forward declaration of `Hamiltonian` instead of including `energy.h`

src/units.h
```diff
@@ -65,44 +68,65 @@ namespace Faunus {
...
-     * Energy             | Thermal energy (kT)
-     * Temperature        | Kelvin (K)
-     * Length             | Angstrom (A)
-     * Charge             | Electron unit charge (e)
-     * Dipole moment      | Electron angstrom (eA)
-     * Concentration      | Particles / angstrom^3
-     * Pressure           | Particles / angstrom^3
-     * Angle              | Radians
+     * energy             | thermal energy unit (kT)
+     * temperature        | kelvin (K)
```
mlund, line 72: Kelvin, Ångstrom, Dalton should start with capital letters. For mass, I prefer "grams per mole". Also, from Wikipedia on Dalton:

_2019 redefinition of the SI base units[edit]: 
The definition of the dalton was not affected by the 2019 redefinition of SI base units,[28][29][1] that is, 1 Da in the SI is still 1/12 of the mass of a carbon-12 atom, a quantity that must be determined experimentally in terms of SI units. However, the definition of a mole was changed to be the amount of substance consisting of exactly 6.02214076×1023 entities and the definition of the kilogram was changed as well. As a consequence, the molar mass constant is no longer exactly 1 g/mol, meaning that the number of grams in the mass of one mole of any substance is no longer exactly equal to the number of daltons in its average molecular mass.[30]_

examples/langevin/montecarlo_lj.yml
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env yason.py
```
mlund, line 1: In a comment block, add info about what this example illustrates

examples/langevin/compare.py
```diff
@@ -0,0 +1,20 @@
...
+r_mc, g_mc = np.loadtxt('rdf_mc.dat', unpack=True, usecols=(0,1))
+r_ld, g_ld = np.loadtxt('rdf_ld.dat', unpack=True, usecols=(0,1))
+
+stats = ks_2samp(r_mc, r_ld)
+if stats[1] > 0.99:
+   print('p-value = {} (no significant difference)'.format(stats[1]))
+else:
+   print('p-value = {} (significant difference)'.format(stats[1]))
+
+plt.plot(r_mc, g_mc, label='Monte Carlo', linewidth=10, alpha=0.5)
```
mlund, line 15: Very nice check to perform during the test suite, i.e. `ctest`. The plotting step should be skipped.

docs/_docs/langevin.md
```diff
@@ -0,0 +1,49 @@
+<script type="text/x-mathjax-config">
+MathJax.Hub.Config({
+  tex2jax: {inlineMath: [['$','$'], ['\\(','\\)']]}
+});
+</script>
+<script src="https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.0/MathJax.js?config=TeX-AMS-MML_HTMLorMML" type="text/javascript"></script>
+
+# Deterministic and Stochastic Dynamics
+The possibility of conducting hybrid molecular dynamics - Monte Carlo (MDMC) schemes is possible by addition of the "Langevin move" in the `moves` section.
```
mlund, line 9: - Rephrase. "possibility" and "possible" has similar meaning
- Give a yaml example
- Insert space after header.
- As we don't have anything else at the moment, perhaps just a single header "Langevin Dynamics"

docs/_docs/langevin.md
```diff
@@ -0,0 +1,49 @@
...
+ABOBA           | [Symmetric Langevin Position-Verlet](http://doi.org/ggrnfs)
+BAOAB           | [Symmetric Langevin Velocity-Verlet](http://doi.org/ggrnfs)
+OABAO           | [Bussi-Parrinello Langevin Position-Verlet](http://doi.org/cjpdjx)
+OBABO           | [Bussi-Parrinello Langevin Velocity-Verlet](http://doi.org/cjpdjx)
+
+We recommend the usage of the splitting scheme "BAOAB" (Symmetric Langevin Velocity-Verlet) due to it being less errorprone with increasing timestep [Leimkuhler & Matthews, pp. 279-281](http://doi.org/dx7v).
+
+For examples of the usage of the Langevin move and combined usage with Monte Carlo moves see the [molecular dynamics](https://github.com/mlund/faunus/blob/master/examples/moleculardynamics) directory.
+
+## Molecular dynamics
```
mlund, line 48: Remove for now?

examples/langevin/montecarlo_lj.yml
```diff
@@ -0,0 +1,28 @@
...
+energy:
+    - nonbonded_coulomblj:
+        coulomb: {epsr: 80, type: plain}
```
mlund, line 4: skip coulomb since the particle charge is zero?

src/forcemove.cpp
```diff
@@ -0,0 +1,172 @@
...
+ * @return a mean square thermal speed in ångström squared per picosecond squared
+ */
+template<typename T=double> static inline T meanSquareSpeedComponent(T mass) {
+    return (pc::kT() / mass * 1.0_kg) * ((1.0_m * 1.0_m) / (1.0_s * 1.0_s));
+}
+
+/**
+ * @brief Generate standard normal random vector
+ */
+class NormalRandomVector {
```
mlund, line 23: Class declaration should be in header file

src/forcemove.cpp
```diff
@@ -0,0 +1,172 @@
...
+/**
+ * @brief Compute a single dimension contribution to a mean square thermal speed of a particle, i.e., compute 〈v²_x〉.
+ *
+ * This is equivalent to a one third of a total mean square thermal speed of a particle in three dimensions,
+ * i.e., 〈v²〉 = 3 × 〈v²_x〉.
+ *
+ * @param mass particle's mass in dalton
+ * @return a mean square thermal speed in ångström squared per picosecond squared
+ */
+template<typename T=double> static inline T meanSquareSpeedComponent(T mass) {
```
mlund, line 16: Does this need to be templated? T=Double should be sufficient.

src/forcemove.cpp
```diff
@@ -0,0 +1,172 @@
...
+// =============== IntegratorBase  ===============
+
+IntegratorBase::IntegratorBase(Space &spc, Energy::Energybase &energy) : spc(spc), energy(energy) {}
+
+void from_json(const json &j, IntegratorBase &i) { i.from_json(j); }
+void to_json(json &j, const IntegratorBase &i) { i.to_json(j); }
+
+// =============== LangevinVelocityVerlet ===============
+
+LangevinVelocityVerlet::LangevinVelocityVerlet(Space &spc, Energy::Energybase &energy, float dt, float gamma)
```
mlund, line 46: float --> double

src/forcemove.h
```diff
@@ -0,0 +1,98 @@
...
+void to_json(json &j, const IntegratorBase &i);
+
+/**
+ * @brief Symmetric Langevin velocity-Verlet method (BAOAB)
+ *
+ * The integrator can conduct normal velocity-Verlet, when gamma = 0.
+ */
+class LangevinVelocityVerlet : public IntegratorBase {
+    float dt;    //!< integration time step (picoseconds)
+    float gamma; //!< friction coefficient (inverse picoseconds)
```
mlund, line 42: --> `double friction_coefficient`

src/forcemove.h
```diff
@@ -0,0 +1,98 @@
...
+
+    //! increment particle's position in an A semi-step
+    inline Point positionIncrement(const Point& velocity);
+    //! increment particle's velocity in a B semi-step
+    inline Point velocityIncrement(const Point& force, const double mass);
+    //! apply fluctuation-dissipation to particle's velocity by solving of Ornstein-Uhlenbeck process in an O-step
+    inline Point velocityFluctuationDissipation(const Point& velocity, const double mass);
+  public:
+    LangevinVelocityVerlet(Space &spc, Energy::Energybase &energy, float dt = 1.0, float gamma = 0.0);
```
mlund, line 51: dt -> time_step and gamma -> friction_coefficient

examples/langevin/compare.py
SHervoe-Hansen: Before we go on with this step, I want to conduct some noise to one of the distributions to check how much difference there would have to be before the [Kolmogorov-Smirnov test](https://en.wikipedia.org/wiki/Kolmogorov%E2%80%93Smirnov_test) consider the the two distributions significantly different given a significance level of 99%.

docs/_docs/langevin.md
SHervoe-Hansen: Possible include til line 19 and then give a yaml example?

src/energy.h
```diff
@@ -1184,6 +1180,7 @@ class Hamiltonian : public Energybase, public BasePointerVector<Energybase> {
     double maxenergy = pc::infty; //!< Maximum allowed energy change
     void to_json(json &j) const override;
     void addEwald(const json &j, Space &spc); //!< Adds an instance of reciprocal space Ewald energies (if appropriate)
+    void force(std::vector<Point> &) override;
```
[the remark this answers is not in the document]
rc83: done

src/energy.h
```diff
@@ -1184,6 +1180,7 @@ class Hamiltonian : public Energybase, public BasePointerVector<Energybase> {
...
   public:
     Hamiltonian(Space &spc, const json &j);
```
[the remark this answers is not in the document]
rc83: done

src/energy.cpp
```diff
@@ -694,6 +695,12 @@ void Hamiltonian::addEwald(const json &j, Space &spc) {
     }
 }
 
+void Hamiltonian::force(std::vector<Point> &forces) {
```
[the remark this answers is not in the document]
rc83: done

src/externalpotential.h
```diff
@@ -27,7 +27,7 @@ class Energybase {
     virtual void to_json(json &j) const;                  //!< json output
     virtual void sync(Energybase *, Change &);
     virtual void init();                               //!< reset and initialize
-    virtual inline void force(std::vector<Point> &){}; // update forces on all particles
+    virtual inline void force(std::vector<Point> &){}; //!< update forces on all particles
```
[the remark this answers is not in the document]
rc83: done

src/forcemove.cpp
```diff
@@ -0,0 +1,172 @@
+#include "forcemove.h"
+#include "random.h"
+namespace Faunus {
+
+namespace Move {
+
+/**
+ * @brief Compute a single dimension contribution to a mean square thermal speed of a particle, i.e., compute 〈v²_x〉.
```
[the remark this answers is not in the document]
rc83: done

src/forcemove.h
rc83: done

src/forcemove.h
rc83: As the method is purely virtual, it has no definition in cpp file.

src/forcemove.cpp
```diff
@@ -0,0 +1,172 @@
...
+ *
+ * @param mass particle's mass in dalton
+ * @return a mean square thermal speed in ångström squared per picosecond squared
+ */
+template<typename T=double> static inline T meanSquareSpeedComponent(T mass) {
+    return (pc::kT() / mass * 1.0_kg) * ((1.0_m * 1.0_m) / (1.0_s * 1.0_s));
+}
+
+/**
+ * @brief Generate standard normal random vector
```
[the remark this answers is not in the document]
rc83: Changed. Yes, it shall be random both in the direction and magnitude.

src/forcemove.cpp
```diff
@@ -0,0 +1,172 @@
...
+ */
+class NormalRandomVector {
+    std::normal_distribution<double> normal_distribution;
+
+  public:
+    NormalRandomVector(double mean = 0.0, double stddev = 1.0) : normal_distribution(mean, stddev){};
+    inline Point operator()() {
+        return Point(normal_distribution(random.engine),
```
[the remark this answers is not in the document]
rc83: done

src/forcemove.cpp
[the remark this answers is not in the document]
rc83: Changed. Yes, it shall be random both in the direction and magnitude.

src/forcemove.cpp
```diff
@@ -0,0 +1,172 @@
...
+        auto &velocity = velocities[i];
+        velocity += velocityIncrement(forces[i], mass);            // B step
+        position += positionIncrement(velocity);                   // A step
+        velocity = velocityFluctuationDissipation(velocity, mass); // O step
+        position += positionIncrement(velocity);                   // A step
+        spc.geo.boundary(position);
+    }
+    // Update forces. The force vector must be initialized with zeros as the resulting force is computed
+    // additively (+=).
+    std::fill(forces.begin(), forces.end(), Point(0, 0, 0));
```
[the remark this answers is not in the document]
rc83: done

src/forcemove.h
rc83: Done & done. However, we should decide how `dt` and `gamma` shall be represented in json. I consider more descriptive names like `time_step` and `friction` or `friction_coef` better.

src/forcemove.h
rc83: Here, the doxygen comment before the declaration helps keeping the rhythm when reading through the header file, as other comments in the same section are longer and before the methods' declarations.

Parameter names in method declarations. I am slowly building an opinion to rather keep them unless obvious and distracting (e.g., in `to_json` and `from_json` methods). It is true that the parameters' names in declarations have no purpose (are redundant) from the compiler's point of view. You also have to make an extra effort to keep them in accord with methods' definitions. However the advantage of more readable code outweighs this tiny need of extra care. Compare for example `void step(PointVector &, PointVector &)` and `void step(PointVector &velocities, PointVector &forces)`. In the former case, you have to jump to the definition to see how to use the method which is inconvenient at least. Are we following any particular style guide in this matter? Here are some additional arguments to support my (still weak) opinion on this.
https://stackoverflow.com/questions/7891526/should-function-declarations-include-parameter-names
https://stackoverflow.com/questions/771492/c-style-convention-parameter-names-within-class-declaration

src/forcemove_test.h
rc83: done

src/forcemove.h
rc83: done

src/forcemove.h
rc83: done

src/forcemove.cpp
rc83: done

src/forcemove.cpp
```diff
@@ -0,0 +1,172 @@
...
+namespace Move {
+
+/**
+ * @brief Compute a single dimension contribution to a mean square thermal speed of a particle, i.e., compute 〈v²_x〉.
+ *
+ * This is equivalent to a one third of a total mean square thermal speed of a particle in three dimensions,
+ * i.e., 〈v²〉 = 3 × 〈v²_x〉.
+ *
+ * @param mass particle's mass in dalton
+ * @return a mean square thermal speed in ångström squared per picosecond squared
```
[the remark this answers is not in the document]
rc83: done

src/forcemove.h
rc83: Nope, done. Only obvious variable names (`spc`) removed.

src/forcemove.h
mlund: I would also prefer more elaborate names, although Faunus input is generally not so. Here I prefer `time_step` and `friction`. Ping @SHervoe-Hansen

src/forcemove.h
mlund: To support your case, I actually think CLion signature refactoring could get confused if names are missing in the declaration. I may be wrong, or perhaps it's a matter of settings. In any case, I don't have strong opinions about this, but have seen some cleaning tools complaining about names in headers.

src/forcemove.h
SHervoe-Hansen: I would personally opt for `time_step` and `friction_coef ` (most common) or `collision_feq` (second most common, but yields the most physical understanding of the parameter). I would avoid only using `friction` as keyword would give very little indication what it does if novice to the method.

mlund: Without any solid linguistic backing, I would have thought it should be "coeff" (?). I would prefer not having abbreviations in the input and `friction` should be pretty telling. I suppose it shouldn't be confused with "coefficient of friction" which, if I understand it correctly, is a unit-less quantity.

SHervoe-Hansen: Honestly I think it would be fine to call it `friction` for keyword, the documentation will obviously have the full written names to explain what property it is!

src/units.h
rc83: The easier part first. The unit names like kelvin, ångström, and dalton should start with a lowercase letter, unlike their abbreviations K, Å, and Da that start with an uppercase letter. A proof: https://goldbook.iupac.org/terms/view/K03374. You know, I am a big fan of typography hence I pay disproportionally huge attention to these tiny details. ;)

It is shame that the dalton was not redefined accordingly. Then g/mol appears to be the natural unit of choice instead. However, I struggle mentally a bit to use intensive units (per mole) to extensive quantities like mass or energy (we have already done it in the `kJmol` unit) as it leads to count in moles of moles. Nevertheless, it is only internal Faunus issue without any outside impact hence it can be resolved easily later. For now, I will use gram per mole as the internal unit of mass.

src/forcemove.cpp
```diff
@@ -0,0 +1,172 @@
...
+
+namespace Move {
+
+/**
+ * @brief Compute a single dimension contribution to a mean square thermal speed of a particle, i.e., compute 〈v²_x〉.
+ *
+ * This is equivalent to a one third of a total mean square thermal speed of a particle in three dimensions,
+ * i.e., 〈v²〉 = 3 × 〈v²_x〉.
+ *
+ * @param mass particle's mass in dalton
```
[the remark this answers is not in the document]
rc83: done

src/forcemove.h
rc83: The keywords are renamed to `time_step` and `friction`.

src/forcemove.cpp
rc83: Un-templated and documentation altered.

A side note for templates: Their definitions have to be available in the translation unit they are used in. Putting a static function template in a cpp file thus makes a perfect sense.

src/forcemove.cpp
```diff
@@ -0,0 +1,172 @@
...
+  public:
+    NormalRandomVector(double mean = 0.0, double stddev = 1.0) : normal_distribution(mean, stddev){};
+    inline Point operator()() {
+        return Point(normal_distribution(random.engine),
+                     normal_distribution(random.engine),
+                     normal_distribution(random.engine));
+    };
+};
+
+NormalRandomVector random_vector; //!< standard normal random vector generator
```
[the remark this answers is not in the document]
rc83: It was some remainder. Fixed.

src/forcemove.cpp
rc83: Moved. Before it was only a “private” class to be used within the translation unit. Looking forward to modules that are introduced in C++20.

src/forcemove.cpp
```diff
@@ -0,0 +1,172 @@
...
+        velocities[i] = random_vector() * std::sqrt(meanSquareSpeedComponent(mass));
+        forces[i] = {0, 0, 0};
+    }
+}
+
+// =============== LangevinMove ===============
+
+LangevinDynamics::LangevinDynamics(Space &spc, std::shared_ptr<IntegratorBase> integrator, unsigned int nsteps)
+    : ForceMoveBase::ForceMoveBase(spc, integrator, nsteps) {
+    name = "ld";
```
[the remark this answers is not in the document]
rc83: Renamed to `langevin_dynamics`; `molecular_dynamics` without “dynamics” would sound odd.

src/forcemove.h
rc83: Done in a short introductory docblock.

examples/langevin/montecarlo_lj.yml
rc83: Done in both langevin and montecarlo examples.

docs/_docs/langevin.md
rc83: MD completely removed from the manual for now. The markup was also fixed to compile to PDF.

src/forcemove.h
rc83: I will create a ticket to resolve this as a separate issue.

rc83:
I keep commits as breadcrumbs to easily accommodate any future change requests. I will squash them together and force-push, when approved.

mlund:
Besides the failing unittest, merge at will!

PR #300, opened 2020-07-03
Fix speciation check for full or empty groups

mlund:
This fixes a bug where products or reactants could be wrongly (de)activated even if the groups are full/empty. This is a problem important only when simulating at extreme conditions where the groups are nearly empty or nearly full. The issue could appear as a non-zero net-charge upon restarting an equilibration run which would typically be initiated with all molecules in their active state. The issue would not appear with swap moves and only if multiple products or reactants were involved in the reaction.


mlund:
@rc83 cherry pick this commit into a v2.4.2 bug fix release?

rc83:
I have created and pushed branch called release-2.4. It may be tagged as v2.4.2 if all test pass. – I have few tests failing locally but that can be different issue.

PR #315, opened 2020-07-23
Refactor structure loading policy

mlund:
Changes to structure loading functionality with.

- io::loadStructure:
  - signature change
  - exception policy change
  - silent exception replaced with conditional

- split a few functions in molecule.h into declaration and definition

- renamed typedef

### DIFF
--- a/src/molecule.cpp
+++ b/src/molecule.cpp
@@ -3,18 +3,16 @@
 #include "geometry.h"
 #include "rotate.h"
 #include "bonds.h"
-#include <fstream>
-#include <iostream>
 #include "spdlog/spdlog.h"
 #include "aux/eigensupport.h"
 
 namespace Faunus {
 
 // global instances available throughout Faunus.
-std::vector<MoleculeData> molecules;
-std::vector<ReactionData> reactions;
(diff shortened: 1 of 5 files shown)


src/molecule.cpp
```diff
@@ -375,15 +369,17 @@ void MoleculeStructureReader::readArray(ParticleVector &particles, const json &j
     }
 }
 
-void MoleculeStructureReader::readFasta(ParticleVector &particles, const json &j_fasta) {
-    if (j_fasta.find("fasta") == j_fasta.end()) {
+void MoleculeStructureReader::readFasta(ParticleVector &particles, const json &input) {
+    if (auto it = input.find("fasta"); it == input.end()) {
         throw ConfigurationError("invalid FASTA format");
+    } else {
```
rc83, line 375: As `it` is accessible in the whole method anyway, the if-init statement oddly introducing the else block is not really needed.

```
auto it = …;
if (it == input.end()) {
    throw …;
}
// useful code
```

mlund: Thanks for the review! I'm trying to develop a habit of reducing variable scope using C++17's if-statement with an initialiser. But I agree it doesn't add a lot of safety here.

PR #327, opened 2020-09-11
Allow cuboids in scattering analysis

mlund:
This allow non-cubic systems in scattering analysis. This is needed for future implementation if scattering calculation in e.g. hexagonal geometries mapped onto a cuboid.


rc83 (approved):

mlund:
ping @bjornstenqvist Does this looks OK for IPBC?

bjornstenqvist (commented):
Looks good to me!

PR #330, opened 2020-09-28
Fix VirtualVolumeAnalysis (Space is a reference)

rc83:
Well, that little ampersand, you know. ;)


mlund (commented):
Thanks for finding this well concealed yet mortal bug! Would it make sense to also here include the virtualvolume analysis to one of the existing test cases (`polymers`, for example)? Alternatively, a tiny two-particle LJ system where the exact result would be known (ould also be useful for many other analysis functions).

PR #335, opened 2020-10-22
Binary search for bond search in Bonded energy

mlund:
When only a subset of particles are moved, the previous algorithm would be of complexity O(N^2). For large polymers this could be a significant bottle-neck. Here we exploit that the list of changed particle index are _sorted_, allowing for binary search. 10-100 times speedup has been observed for large polymers.


rc83 (approved):
I have just deleted my review-in-progress as all comments were fixed in the recent commits. ;) I will try to nicely squash merge it if my permissions allow.

PR #337, opened 2020-11-26
Add support for Trompeloeil mocking framework

rc83:
https://github.com/rollbear/trompeloeil

# Description

Allows better unit testing with doctest by creating mock objects and asserting their methods calls.


mlund (approved):
Nice!

PR #338, opened 2020-11-26
Celllist preparation

rc83:
# Description

Various fixes, refactorizations and improvements done during cell list programming. They should not raise any emotions. To be reviewed and merged per commit basis, i.e., **“Rebase and merge”**.

The codebase should be gradually cleaned from C-style idioms and non-idioms. 

* A function performing some task should not return bool to indicate success or failure, but rather only throw an exception upon failure. That avoids illogical ifs – the code does a task and does not answer a yes-no question. Such ifs also tend to be nested creating difficult-to-read blocks. Using exceptions on the other hand leads to a nicely readable liner code. 
* The construct `if(condition) return true; else return false;` is unnecessary verbose: plain `return condition;` does the same.
* For discussion: Changing the code style for references (and pointers) by moving the symbol from the variable name to the type, e.g., `MyClass &object` → `MyClass& object`, `MyClass &` → `MyClass&`. C is value oriented with a rather simple type system in contrast to C++ that strongly relies on (multiple) inheritance and uses templates to control types. Hence the first style is natural for C, but not for C++ as `MyClass` and `MyClass&` are different _types_. The more I am involved in metaprogramming (templates) the more I consider C style rather distracting in C++.

### DIFF
--- a/src/speciation.cpp
+++ b/src/speciation.cpp
@@ -332,7 +332,7 @@ bool SpeciationMove::deactivateAllReactants(Change &change) {
     for (auto [molid, N_delete] : molecular_reactants) { // Delete
         if (N_delete <= 0 or Faunus::molecules[molid].isImplicit()) {
             continue;                         // implicit molecules are added/deleted after move
-        } else if (molecules[molid].atomic) { // reactant is an atomic group
+        } else if (Faunus::molecules[molid].atomic) { // reactant is an atomic group
             auto mollist = spc.findMolecules(molid, Tspace::ALL);
             assert(range_size(mollist) == 1);
             auto target = spc.findMolecules(molid, Tspace::ALL).begin();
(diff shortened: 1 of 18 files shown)


src/speciation.cpp
```diff
@@ -387,57 +394,52 @@ bool SpeciationMove::deactivateAllReactants(Change &change) {
...
-        return true;
-    }
+    return std::find_if(molecular_reactants.begin(), molecular_reactants.end(), isExhausted) ==
+           molecular_reactants.end(); // no molecule exhausted?
 }
 
 void SpeciationMove::_move(Change &change) {
     assert(other_spc != nullptr);        // knowledge of other space should be provided by now
-    if (not Faunus::reactions.empty()) { // global list of reactions
+    try {
```
mlund, line 425: Have you checked that all tests pass? The boolean return type is used to catch a number of things (I can't remember them all) that could influence the speciation flow. I like the idea of throwing much more so we can get rid of original comments like "slip out the back door" :-)

mlund (approved):
Nice cleanup! As you write, it's not provocative and I recommend merging as is. Please double-check the speciation checks and verify that no logic has been affected by the replacement of bools with exceptions (see single comment).

PR #339, opened 2020-12-17
Fix issues with a few recent commits

mlund:
- cluster move output: include monomers in distribution
- update temper move to updated movebase (req. mpi=on during
  compilation)
- trivial renaming of variables in volumemove (readability)


mlund:
ping @MarcoPolimeni

rc83 (approved):
Nice clean up. I will merge with a more descriptive commit message.

BTW I tweaked the exception hierarchy a bit yesterday that `ConfigurationError` can now hold a usage tip directly. The usage tip thus travels together with the exception and can be logged in the same moment.

PR #341, opened 2021-01-11
Use Conan for dependency handling

mlund:
This PR implements Conan (https://conan.io) dependency handling in the CMake build chain to replace the manual download/build steps currently used. See the updated `install.md` for build details.

- [x] test on macOS x86
- [ ]  test on linux


mlund:
We will instead use CPM (#380)

PR #345, opened 2021-02-04
Simplify JSON processing in from_json

rc83:
# Description

Brief and definitely not comprehensive refactorisation to unify and simplify error handling when reading JSON.

- Move commonly repeated patterns from `from_json` to functions `jsonSingleItem`, `findAtomByName` and `findMoleculeByName`. Error states are handled with exceptions.
- Use `ConfigurationError` to indicate JSON error in `from_json`. Improve text and flow of error messages. JSON snippet can be attached to `ConfigurationError` and shown at debug level.
- Refactor routing `from_json` functions for moves, analyses, energies, bonds and reaction coordinates. A single if-chain is followed to simplify the logic and making unknown JSON keys obvious. A request of features not included at compile time is treated as an explicit error.
- Refactor `from_json` methods of moves, analysis, energies, bonds and reaction coordinate.
- Refactor classes in IO namespace to use `findAtomByName`.
- Remove usageTip from the exception message and display it directly on stderr.
- Improve Faunus error message on empty input.
- Improve code style.

Question: Should be allowed to have more key-value objects in the same array item? Currently it is silently supported (except of bonds), however I did not verified if the code correctly handles this case.
Answer: It shall not be allowed.

Remark: With structural binding `const auto& […]` it becomes more obvious that the reference symbol `&` belongs rather to the type than to the variable name. Hence I suggest to change the code style from C-like to C++-like, e.g., `json &j` → `json& j`.
Resolution: C++ style will be followed.

Another commit just improve a compatibility issue with `pyyaml` I have encountered by accident. I think that the `pyyaml` parser does not fully adhere to the standard but the workaround is painless.

Comments welcome. After approval, please let me merge the PR by myself.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -92,7 +92,7 @@ PenaltyBreakString: 1000
 PenaltyBreakTemplateDeclaration: 10
 PenaltyExcessCharacter: 1000000
 PenaltyReturnTypeOnItsOwnLine: 60
-PointerAlignment: Right
+PointerAlignment: Left
 ReflowComments:  true
 SortIncludes:    false
 SortUsingDeclarations: true
@@ -111,8 +111,7 @@ SpacesInContainerLiterals: true
 SpacesInCStyleCastParentheses: false
 SpacesInParentheses: false
 SpacesInSquareBrackets: false
-Standard:        Cpp11
+Standard:        c++17
 TabWidth:        8
 UseTab:          Never
 ...
-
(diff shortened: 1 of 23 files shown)


mlund (approved):

rc83:
I took the refactorisation more seriously hence a list of new commits appeared, mainly to follow the process. I will squash them to few commits upon merging. I have updated the initial description to reflect all changes. If you do not have any critical comments I will merge this PR later today.

mlund:
> I took the refactorisation more seriously hence a list of new commits appeared, mainly to follow the process. I will squash them to few commits upon merging. I have updated the initial description to reflect all changes. If you do not have any critical comments I will merge this PR later today.

Thanks, nice changes; merge at will.

PR #363, opened 2021-05-07
Add force support for `HarmonicTorsion`

mlund:
This adds force support for `HarmonicTorsion` and includes some cleanup of `BondData` and force evaluation.

- Add `bend` example for MD/MC testing of harmonic bonds and angles
- `BondData` refactoring

### DIFF
--- a/src/bonds.cpp
+++ b/src/bonds.cpp
@@ -218,12 +218,12 @@ void HarmonicTorsion::setEnergyFunction(const ParticleVector& particles) {
             Point force1 = -(force0 + force2); // no net force
             return {{indices[0], force0}, {indices[1], force1}, {indices[2], force2}};
         } else { // @todo which is faster?
-            const Point ba = calculateDistance(particles[indices[1]].pos, particles[indices[0]].pos);
-            const Point bc = calculateDistance(particles[indices[1]].pos, particles[indices[2]].pos);
+            const Point ba = calculateDistance(particles[indices[0]].pos, particles[indices[1]].pos); // b->a
+            const Point bc = calculateDistance(particles[indices[2]].pos, particles[indices[1]].pos); // b->c
             const auto inverse_norm_ba = 1.0 / ba.norm();
             const auto inverse_norm_bc = 1.0 / bc.norm();
             const auto angle = std::acos(ba.dot(bc) * inverse_norm_ba * inverse_norm_bc);
-            const auto forcemagnitude = 2.0 * half_force_constant * (angle - equilibrium_angle);
+            const auto forcemagnitude = -2.0 * half_force_constant * (angle - equilibrium_angle);
             const Point plane_babc = ba.cross(bc).eval();
             Point force0 = (forcemagnitude * inverse_norm_ba * ba.cross(plane_babc).normalized());
             Point force2 = (forcemagnitude * inverse_norm_bc * -bc.cross(plane_babc).normalized());
--- a/src/bonds.cpp
+++ b/src/bonds.cpp
@@ -201,35 +201,35 @@ void HarmonicTorsion::setEnergyFunction(const ParticleVector& particles) {
         return half_force_constant * delta_angle * delta_angle;
     };
     forceFunc = [&](Geometry::DistanceFunction calculateDistance) -> std::vector<IndexAndForce> {
-        if constexpr (true) {
-            // see https://github.com/OpenMD/OpenMD/blob/master/src/primitives/Bend.cpp
-            Point vec1 = calculateDistance(particles[indices[0]].pos, particles[indices[1]].pos);
-            Point vec2 = calculateDistance(particles[indices[2]].pos, particles[indices[1]].pos);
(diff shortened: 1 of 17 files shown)


src/bonds.cpp
```diff
@@ -200,6 +200,29 @@ void HarmonicTorsion::setEnergyFunction(const ParticleVector& particles) {
...
+            cosine_angle = 1.0;
+        } else if (cosine_angle < -1.0) {
+            cosine_angle = -1.0;
+        }
+        auto sine_angle = std::sqrt(1.0 - cosine_angle * cosine_angle);
+        if (std::fabs(sine_angle) < 1e-6) {
+            sine_angle = std::numeric_limits<double>::epsilon();
+        }
+        const auto angle = std::acos(cosine_angle);
+        const auto du_dangle = 2.0 * half_force_constant * (angle - equilibrium_angle);
```
SHervoe-Hansen: minus sign missing.
F = - dU/dr

src/bonds.cpp
```diff
@@ -200,6 +200,29 @@ void HarmonicTorsion::setEnergyFunction(const ParticleVector& particles) {
...
+        auto force0 = du_dangle / sine_angle * d01_inv * (r21 * d21_inv - r01 * d01_inv * cosine_angle);
```
SHervoe-Hansen: In more Eigen friendly manner we could maybe write it as:

force0 = du_dangle / [ r01 * norm(10x(10x12)) ]

src/bonds.cpp
```diff
@@ -200,6 +200,29 @@ void HarmonicTorsion::setEnergyFunction(const ParticleVector& particles) {
...
+        auto force2 = du_dangle / sine_angle * d21_inv * (r01 * d01_inv - r21 * d21_inv * cosine_angle);
```
SHervoe-Hansen: force2 = du_dangle / [ r12 * norm(21x(10x12)) ]

SHervoe-Hansen (commented):
It seems there is a sign error that cancels out if all tests passes.

src/bonds.cpp
SHervoe-Hansen: Note the cross product of vector 10 and 12 is used in both expressions so maybe calc that first and take the norm of the cross product of between the newly calculated vector and the vector required.

src/bonds.cpp
mlund: I think it's just the direction of the vector (01 vs 10). I checked flipping the sign, and see large differences compared to MC.

src/bonds.cpp
mlund: This might be faster as it uses one less trigonometric function (sqrt) in exchange of multiplications. However, your suggestion gives NaN and `ctest -R unittests -V` fails. Would you be willing to double-check with the current code?

SHervoe-Hansen:
It passes the tests in terms of RDF is the same for Langevin dynamics and Monte Carlo and does not return NaN.

bend_rdf

mlund:
> It passes the tests in terms of RDF is the same for Langevin dynamics and Monte Carlo and does not return NaN.
> 
> bend_rdf

Thanks, the cross-product code works in Debug mode only, but fails on both GCC and Clang if with RelWithDebMode. I tried updating Eigen from 3.3.7 --> 3.3.9 but to no avail. I kept the code, but enclosed it in a `constexpr(false)` which will be evaluated at compile-time for further testing.

mlund:
@SHervoe-Hansen in the unittest,`bonds.cpp:312`, the distance function is defined as `b - a`, whereas in `Cuboid::vdist()` (and all other geometries) it's `a - b`. Is there a lucky cancellation of signs somewhere?

mlund:
> I have been testing for more odd behaviour, it turns out the debug only code can run in release mode if you insert std::cout << force0; and std::cout << force2 before the calculation of force1, which I can only associate with uninitialised variable behaviour.

I think we just ran into one of Eigen's [common pitfalls](https://eigen.tuxfamily.org/dox/TopicPitfalls.html) :-) Fixed in latest commit. Still a bit worried about the force sign more generally in Faunus (bonds, pair potential).

SHervoe-Hansen:
The commit https://github.com/mlund/faunus/pull/363/commits/4e34050d54e6552a20e700324328e32bfe49e56e now reflects the addition of a unit test checking the direction of the force vectors to be correct. The unit test system is illustrated here.
AngularUnitTest

The angular harmonic potential now passes this test.

mlund:
> The angular harmonic potential now passes this test. Agreement between MC and MD however still remains to be tested in addition to the speed of the two implementations.

I tried `ctest -R bend` with the latest "cross product" variant and got the graph below (green), indicating a sign is flipped somewhere. You can switch version by simple changing the `bool` in `if constexpr()`; no need to uncomment.

Screenshot 2021-05-12 at 23 05 44

SHervoe-Hansen (approved):
The code and physics now match and passes all tests. Ready for merge.

PR #369, opened 2021-05-25
PCG random number generator

mlund:
This is an optional drop-in replacement for C++'s Mersenne Twister. See https://www.pcg-random.org/ for more information.


docs/_docs/install.md
```diff
@@ -74,6 +74,7 @@ CMake Option                         | Description
 `-DENABLE_PYTHON=OFF`                | Build python bindings (experimental)
 `-DENABLE_FREESASA=ON`               | Enable SASA routines (external download)
 `-DENABLE_TBB=OFF`                   | Build with Intel Threading Building Blocks (experimental)
+`-DENABLE_PCG=OFF`                   | Use PCG random number generator instead of C++'s Mersenne Twister
```
SHervoe-Hansen, line 77: Maybe we should keep it on as default due to PCG superiority over Mersenne Twister?

examples/cluster-ideal/cluster-ideal.yml
```diff
@@ -5,7 +5,7 @@
 #     N_c = 1 + 4 * pi * Rc**3 * N / (3 * L**3) = 1.22 in this case
 #
 temperature: 300
-random: {seed: fixed}
+random: {seed: hardware}
```
SHervoe-Hansen, line 8: Any requirements in terms of our examples and seed?

SHervoe-Hansen (commented):
The PCG RNG looks amazing from the material linked in the PR.

However given that mind I think we should set the value to default and perhaps also update the documentation in case one attempts to reproduce results given a specific RNG.

docs/_docs/install.md
mlund: I agree it should be better but we currently don't know so much about it. I think we can leave it on by default after the next release.

examples/cluster-ideal/cluster-ideal.yml
mlund: I think the cluster move displaced too far as `dp=20` is exactly half the box lengtht. I have restored `fixed` and instead set `dp=19.9999` which passed. Otherwise we would randomly get mass center out of sync.

mlund:
> Furthermore multiple tests returned
> `could not initialize random engine - falling back to fixed seed.`
> Is this related to the PCG?

Yes, the state is specific to the engine, but I think the message explanation is OK'ish. If we set PCG to default we should update (or remove) the state-info from the state files.

PR #375, opened 2021-06-16
Xyzfile syntax check

mlund:
More meaningful output and syntax check for XYZ file loader.
@SHervoe-Hansen in response to Coralie's issue.


SHervoe-Hansen (approved):
Looks great and error handling is spot on! Merge as you please and I think Coralie will also be happy.

PR #397, opened 2021-11-01
Sasa celllist

Smutekj:
# Description

Added cell list neighbour search for sasa calculation in cuboid and sphere geometries

## Checklist

- [x] `make test` passes with no errors
- [x] the source code is well documented
- [x] new functionality includes unittests
- [x] the user-manual has been updated to reflect the changes
- [x] I have installed the provided commit hook to auto-format commits (requires `clang-format`):

  ``` bash
  ./scripts/git-pre-commit-format install
  ```

- [x] code naming scheme follows the convention:

  Style        | Elements
  ------------ | ---------------------
  `PascalCase` | classes, namespaces
  `camelCase`  | functions
  `snake_case` | variables


mlund:
@Smutekj There seems to be conflicts, so I think you'll need to merge `master` with `sasa-celllist` locally, then commit again. Something like this:
``` bash
git pull
git checkout sasa-celllist
git merge master
```

src/sasa.cpp
```diff
@@ -233,4 +239,208 @@ TEST_CASE("[Faunus] SASAPBC") {
...
+                for (const auto changed_index : changed_atom_indices) {
+                    cell_list->updateMemberAt(changed_index,
+                                              spc.particles.at(changed_index).pos + spc.geometry.getLength() / 2.);
+                }
+            }
+        }
+    }
+}
+
+template <typename CellList_T> void SASACellList<CellList_T>::createCellList(Space& spc) {
```
mlund, line 351: slightly more general and less dependent if this function takes a pair of templated iterators (begin, end) and the geometry.

src/celllistimpl.h
```diff
@@ -765,6 +775,17 @@ template <class TBase> class CellListReverseMap : public TBase {
         this->update(member, this->index(new_cell_coordinates));
     }
 
+    /**
+     * @brief returns true if member is present in the cell list false if not
+     * @param member
+     */
+    bool containsMember(const Member& member) {
+        if (member2cell.count(member) == 0) {
```
mlund, line 783: or simpler `return member2cell.count(member) != 0;`

src/celllistimpl.h
```diff
@@ -290,8 +290,18 @@ class PeriodicBoundaryGrid : public GridBase<TGridType>, virtual public Abstract
      * @return  cell index
      */
     CellIndex index(const CellCoord& coordinates) const override {
-        const auto pbc_coord = (coordinates - (coordinates / this->getCellListEnd()) * this->getCellListEnd()).eval();
-        return Base::index(pbc_coord);
+        // const auto pbc_coord = (coordinates - (coordinates / this->getCellListEnd()) *
```
mlund, line 293: remove outcommented code

src/energy.cpp
```diff
@@ -1284,7 +1302,7 @@ void SASAEnergyBase::to_json(json& j) const {
     roundJSON(j, 6); // set json output precision
 }
 
-std::vector<size_t> SASAEnergy::findTargetIndices(Change& change) {
+std::vector<size_t> SASAEnergy::findChangedIndices(Change& change) {
```
mlund, line 1305: This function is rather long. Perhaps it could be split in some way?

src/sasa.cpp
```diff
@@ -233,4 +239,208 @@ TEST_CASE("[Faunus] SASAPBC") {
...
+    cell_offsets.clear();
+    for (auto i = -1; i <= 1; ++i) {
+        for (auto j = -1; j <= 1; ++j) {
+            for (auto k = -1; k <= 1; ++k) {
+                cell_offsets.push_back(CellCoord(i, j, k));
+            }
+        }
+    }
+
+    radii.clear();
```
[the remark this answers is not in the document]
Smutekj: I wanted to use that as you suggest but without the `auto` because radii is already declared as a member, but then the compiler complains:

error: no match for ‘operator=’ (operand types are ‘std::vector<double>’ and ‘ranges::transform_view<ranges::ref_view<std::vector<Faunus::Particle> >, Faunus::SASA::SASA::init(Faunus::Space&)::<lambda(auto:46&)> >’)
  146 |     radii = spc.particles | ranges::cpp20::views::transform([](auto &i){return 0.5 * i.traits().sigma;});

src/celllistimpl.h
```diff
@@ -789,14 +810,14 @@ template <class TBase> class CellListReverseMap : public TBase {
...
-            TBase::update(member, old_cell_index, new_cell_index);
+            TBase::move(member, old_cell_index, new_cell_index);
             member2cell.at(member) = new_cell_index;
         }
     }
 
   private:
     // TODO as a vector because size of Particle vector is constant
-    std::map<Member, CellIndex> member2cell; //!< mapping from a member to its cell
+    std::unordered_map<Member, CellIndex> member2cell; //!< mapping from a member to its cell
```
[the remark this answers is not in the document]
Smutekj: Thought it would for large number of active particles because std::map has O(logN) complexity of insertions and deletions. But I have not found yet a system where this would appear, since it takes only very little time compared to the rest of calculation. I think I will try it with profiling in clion, but I have never used anything like that so it is going to take me a while.

mlund (approved):
Very nice commit with high quality C++ code. I made a few cosmetic changes and approved it. Merge at will and please select **Squash Merge** whereby it will appear as a single commit in the master branch.
One last suggestion is to add doxygen `/** */` blocks to the three classes in `sasa.h` as they are currently undocumented.

src/celllistimpl.h
```diff
@@ -442,12 +444,20 @@ template <typename TMember, typename TIndex> class DenseContainer : virtual publ
         return indices;
     }
 
-    DenseContainer(std::size_t size) { container.resize(size); }
+    DenseContainer(std::size_t size) {
```
Smutekj, line 447: The DenseContainer throws SIGKILL if `size` is larger than cca 1 billion, so I added a safecheck.

src/celllistimpl.h
```diff
@@ -508,7 +518,7 @@ template <typename TMember, typename TIndex> class SparseContainer : virtual pub
   private:
     Index index_end;                    //!< the lowest index not allowed to appear, i.e., the grid size
     const Members empty_set;            //!< an empty set, e.g., for out of boundary conditions
-    std::map<Index, Members> container; //!< container itself
+    std::unordered_map<Index, Members> container; //!< container itself
```
Smutekj, line 521: Tested it and It was faster.

PR #424, opened 2023-02-01
Update minimum cmake version stated in manual

SHervoe-Hansen:
# Description
Cmake versions 3.22 >= x >= 3.20 will all yield the error:
```
CMake Error at /usr/share/cmake/Modules/ExternalProject.cmake:2746 (message):
  URL_HASH is set to

    SHA256=48986b507497f027e4fa1144a08c2d0b6d81fb476fad024956f8104448ca9ad8;DOWNLOAD_EXTRACT_TIMESTAMP;true

  but must be ALGO=value where ALGO is

    MD5|SHA1|SHA224|SHA256|SHA384|SHA512|SHA3_224|SHA3_256|SHA3_384|SHA3_512

  and value is a hex string.
```

### DIFF
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -20,9 +20,26 @@
         <rule entity="LOCAL_VARIABLE" visibility="ANY" specifier="ANY" prefix="" style="SNAKE_CASE" suffix="" />
       </rules>
     </Objective-C-extensions>
-    <clangFormatSettings>
-      <option name="ENABLED" value="true" />
-    </clangFormatSettings>
+    <Objective-C-extensions>
+      <rules>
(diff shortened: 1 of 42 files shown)


mlund (commented):
Thanks for the PR! I'm not able to reproduce the error on my macos system with cmake 3.25.2. Could you describe how the current version fails?

SHervoe-Hansen:
I can reproduce the error on a Linux system with cmake v. 3.20.2 (Faunus manual states minimum  3.16+) and yields the error message first stated.

I have furthermore attempted to update cmake to v. 3.22.1 and it still yields an error.

SHervoe-Hansen:
Okay so I upgraded cmake to 3.25.2 and now it works without errors.

I will change the PR to change the manual so that it states Faunus requires minimum cmake. 3.25.

PR #441, opened 2023-12-12
Tempering update

mlund:


src/move.cpp
```diff
@@ -762,9 +767,12 @@ double ParallelTempering::bias([[maybe_unused]] Change& change, double uold, dou
 
 void ParallelTempering::_accept([[maybe_unused]] Change& change) {
     acceptance_map[partner->getPair(mpi.world)] += 1.0;
+    exchange_map[mpi.world.rank()].push_back(partner->getPair(mpi.world));
+    
 }
 void ParallelTempering::_reject([[maybe_unused]] Change& change) {
     acceptance_map[partner->getPair(mpi.world)] += 0.0;
+    exchange_map[mpi.world.rank()].push_back(std::minmax({-1, -1}));
```
mlund, line 775: Call to `minmax` seems unnecessary

docs/schema.yml
```diff
@@ -849,8 +849,8 @@ properties:
                             default: isotropic
                         format:
                             type: string
-                            enum: [XYZQI, XYZQ, XYZ]
-                            default: XYZQI
+                            enum: [xyzqi, xyzq, xyz]
+                            default: xyzqi
                             description: "Which particle properties to exchange (I=atom id, Q=charge)"
```
mlund, line 854: Make `I` and `Q` lowercase

src/move.cpp
```diff
@@ -698,6 +698,11 @@ void ParallelTempering::_to_json(json& j) const {
         auto id = fmt::format("{} <-> {}", pair.first, pair.second);
         exchange_json[id] = {{"attempts", acceptance.size()}, {"acceptance", acceptance.avg()}};
     }
+    auto& exchange_json2 = j["exchangesmpi"] = json::object();
+    for (const auto& [mpI, exchange] : exchange_map) {
```
mlund, line 702: I guess it's `mpI` due to clash with namespace? Perhaps rename to `_mpi` or alternatively something more verbose.

src/move.h
```diff
@@ -532,6 +532,7 @@ class ParallelTempering : public Move {
     std::unique_ptr<MPI::Partner> partner;     //!< Policy for finding MPI partners
     Geometry::VolumeMethod volume_scaling_method = Geometry::VolumeMethod::ISOTROPIC; //!< How to scale volumes
     std::map<MPI::Partner::PartnerPair, Average<double>> acceptance_map;              //!< Exchange statistics
+    std::map<int, std::vector<MPI::Partner::PartnerPair>> exchange_map;              //!< Exchange statistics
```
mlund, line 535: The memory requirements of the data structure scales with number of steps and could be an issue for long simulations. Perhaps better to simply let each replica stream to a file?

mlund (commented):
Thanks for the PR! It would be great to add the analysis to the `examples/temper/temper.yml` to illustrate how to use it. See specific comment below.

src/move.h
IVinterbladh: That is true. I thought about streaming to a file at first, but did not get it to work.. I will look into it again.

PR #455, opened 2024-07-19
Use optional integer to store exchange partner

mlund:
This should have the exact same output but with slightly better code readability / safety since the non-value -1 is replaced with an optional structure.


IVinterbladh (approved):
Looks good, thanks!

PR #458, opened 2024-10-29
Fix anglescan when the two molecules are different

mlund:
Anglescan would wrongly assume that both molecules are the same. @IVinterbladh could you verify that the bug is fixed?

### DIFF
--- a/src/actions.cpp
+++ b/src/actions.cpp
@@ -50,7 +50,8 @@ AngularScan::AngularScan(const json& j, const Space& spc)
  */
 void AngularScan::Molecule::initialize(const Space::GroupVector& groups, int molecule_index)
 {
-    namespace rv = std::views;
+    using std::views::transform;
+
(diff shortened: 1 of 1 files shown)


src/actions.cpp
```diff
@@ -119,7 +124,7 @@ void AngularScan::operator()(Space& spc, Energy::Hamiltonian& hamiltonian)
 
 #pragma omp parallel for
         for (const auto& q1 : angles.quaternions_1) {
-            auto particles1 = molecules.second.getRotatedReference(spc.groups, q1);
+            auto particles1 = molecules.first.getRotatedReference(spc.groups, q1);
```
mlund, line 127: Source of the problem!

IVinterbladh (approved):
Code is tested and now it works using two different particles for the anglescan method. The anglescan correctly distinguishes the two particles.

PR #477, opened 2026-01-26
Fix MPI deadlock in GibbsMatterMove

mlund:
## Summary

Fixes MPI deadlock when both `moltransrot` and `gibbs_matter` moves are enabled in Gibbs ensemble simulations.

**Root cause:** RNG desynchronization between MPI processes due to `std::uniform_int_distribution` making variable numbers of engine calls for different ranges (rejection sampling). The two cells have different molecule counts, causing `mpi.random` states to diverge, which led to different move selections and eventual deadlock.

**Fix:**
- Use `Faunus::random` (local) for molecule selection instead of `mpi.random`
- Add MPI exchange to synchronize early returns when one process fails

## Test plan

- [x] Verified fix with test case that previously deadlocked
- [x] Multiple successful runs confirm stability

### DIFF
--- a/src/move.cpp
+++ b/src/move.cpp
@@ -768,22 +768,32 @@ void GibbsMatterMove::_from_json(const json& j)
 
 void GibbsMatterMove::_move(Change& change)
 {
-    auto& random_engine = mpi.random;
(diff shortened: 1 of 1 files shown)


IVinterbladh (approved):
Tested and it works!

